PR #124, opened 2026-04-22
Fix NullReferenceException in FindIndexOfNextParaMark

aslakhellesoy:
Hello, and thanks for this great library!

I found a bug where an NRE is thrown when there are certain bookmarks. This PR has a new test that reproduces the bug, and a change to fix it. All other tests are passing.

## Summary

- Fix `NullReferenceException` in `WmlComparer.FindIndexOfNextParaMark` when comparing documents where one has `bookmarkStart`/`bookmarkEnd` as direct children of `w:body` (siblings of `w:p` rather than children)
- The method assumed all elements in the comparison unit array were `ComparisonUnitWord`, but body-level bookmarks produce other `ComparisonUnit` types. The `as ComparisonUnitWord` cast returned `null`, and `.DescendantContentAtoms()` threw
- Added null guards for both the `as` cast (skip non-word units) and the `LastOrDefault()` result

## Test plan

- [x] Added `WmlComparerBodyLevelBookmarkTests` with test documents that reproduce the crash
- [x] Test fails before the fix with `NullReferenceException` at `FindIndexOfNextParaMark`
- [x] Test passes after the fix
- [x] Full test suite passes (1206 passed, 0 failed, 1 skipped)


JSv4:
Thanks for the contribution! Will take a look

PR #259, opened 2026-07-06
Fix WmlComparer header references in redlines

roverw:
Fixes WmlComparer header reference handling in redline output.


JSv4:
Thanks for the PR! Just to be clear on what WmlComparer does - I don't think it actually *compare* header/footer content — your changes stops the redline output from dropping the header/footer references, so the source document's headers and footers are preserved and render again. Probably losing them outright, but if the headers changed between the two versions, that change still won't be tracked.

Actual header/footer comparison is coming in the new DocxDiff engine I've been working on (`CompareHeadersFooters`, on by default there) — that should be released soon (like in a week or two).

I pushed a small follow-up to the branch: a code comment making the carried-not-compared distinction explicit, plus the missing CHANGELOG entry.

roverw:
Thanks - also - great to know about the incoming changes!
The fix is about a broken file, which I used to test the redlining with, that had an image in the header and this image was missing inside the redlined file.

PR #294, opened 2026-07-29
feat(session): footnote/endnote authoring (DocxSession)

### LINKED ISSUE #276: feat(session): footnote/endnote authoring (DocxSession)
Follow-up from #236, which flagged this explicitly: *"Related but **separate**: footnote/endnote authoring (also unsupported) — track independently."*

`DocxSession` can author paragraphs, rules, tables, headers/footers (as of #274), formatting, and annotations — but **not footnotes/endnotes**. The `FootnoteRefNotSupported` `EditErrorCode` is the standing "not yet" placeholder: inserting a `w:footnoteReference` via the markdown subset is explicitly rejected.

## Scope

Anchor-addressed footnote/endnote authoring on `DocxSession`, rippled through all layers per `CLAUDE.md`:

- **Create** a footnote/endnote: add the definition to the `FootnotesPart`/`EndnotesPart` (creating the part + the two Word-reserved separator notes if absent), assign a fresh note id, and insert the body-side `w:footnoteReference`/`w:endnoteReference` run at a character offset in a paragraph.
- **Edit/delete** an existing note: the projection already exposes `fn`/`en` scopes, `GetEditSummary` counts them, and `ReplaceText`/`DeleteBlock` already preserve/handle note refs (see `NoteReferenceNames` / `IsNoteRefOnlyRun` in `DocxSession`).
- Ripple: `DocxSessionOps` → WASM `DocxSessionBridge` + npm `DocxSession` → stdio `Dispatcher` + `docx-scalpel` → tests → `docs/architecture/docx_mutation_api.md`.

## Notes

- Retire `FootnoteRefNotSupported` (or narrow it to "via the markdown subset — use the dedicated op") once the dedicated op lands.
- Undo/redo must reconcile `FootnotesPart`/`EndnotesPart` **part** create/delete — the snapshot-reconcile mechanism #274 added for header/footer parts (`RestoreSnapshot` → `ReconcileHeaderFooterParts`, keyed by relationship id) is the pattern to extend.

JSv4:
## Summary

Closes #276.

`DocxSession` could already **read** notes (the projection's `fn`/`en` scopes, `EditSummary.FootnoteCount`), **edit** one (`ReplaceText` on the note's `p:fn`/`p:en` paragraph) and **delete** one (`DeleteBlock` on the `fn`/`en` definition, which also strips every reference to it). Creation was the only missing verb. This adds exactly that — and no separate edit/delete-note op, because the existing anchor-addressed ops already are those.

### New public surface

```csharp
InsertFootnote(anchorId, characterOffset, markdown)
InsertEndnote(anchorId, characterOffset, markdown)
```

Both delegate to one private `InsertNote(bool isFootnote, …)`, mirroring `SetHeaderText`/`SetFooterText` → `SetHeaderFooterText`.

`Created` returns the note **definition** anchor (kind `fn`/`en`) plus its paragraph anchors (kind `p`, scope `fn`/`en`); `Modified` returns the citing body paragraph. That's the whole lifecycle handoff:

```csharp
var made     = session.InsertFootnote(bodyAnchor, 0, "Source: 2025 annual report.");
var notePara = made.Created.First(a => a.Kind == "p" && a.Scope == "fn").Id;
var noteDef  = made.Created.First(a => a.Kind == "fn").Id;

session.ReplaceText(notePara, "Source: 2026 annual report.");  // edit the note
session.DeleteBlock(noteDef);                                  // drop it + the body citation
```

### What the first note in a document creates

The whole scaffold Word writes, not just the definition:

- the `FootnotesPart`/`EndnotesPart` with Word's two reserved notes (`w:type="separator"` id `-1`, `continuationSeparator` id `0`) — which the projector already filters out of the anchor index and `DeleteBlock` already refuses to remove;
- the `w:footnotePr`/`w:endnotePr` settings declaration, slot-inserted via the shared `EnsureSettingsChildInOrder` so the settings part is **never wholesale reordered** (the discipline `w:evenAndOddHeaders` established in #274);
- find-or-create `FootnoteText`/`EndnoteText` + superscript `FootnoteReference`/`EndnoteReference` styles, so a house style wins and a document that defined neither doesn't render the citation as full-size body text.

A second note reuses all of it and only appends a definition.

### Design calls worth reviewing

- **Note ids are kept ascending in *reference* order** — the invariant every Word-authored document holds, and one renderers depend on (LibreOffice pairs a body marker against the id-sorted definition list, so a first-cited note holding the highest id silently renders the *wrong note text*). A citation following every existing one takes `max(id)+1`, scanning references as well as definitions so non-contiguous ids can't alias an existing note (`DS322`, 1/5/9 fixture); one landing earlier takes the smallest id cited after it and shifts that note and every higher one up by one across all parts, never touching Word-reserved notes (`DS336`/`DS337`/`DS338`). **This was found by smoke-testing a real 94-footnote document — see the smoke-test comment below; it would otherwise have shipped.**
- **The character offset reuses the existing mechanism.** `SplitRunsAtOffset` + `SplitInlineContainersAtOffset` — the same pair `SplitParagraph` and `ApplyFormat` use — so a citation lands cleanly mid-run and inside a hyperlink. No second offset walker to drift.
- **Body paragraphs only.** Word does not allow a note reference inside a header/footer story or inside another note, and the `fn`/`en` scopes are note *definitions*, not citation hosts. A non-body anchor is `AnchorWrongKind` rather than a document Word offers to repair.
- **Undo/redo** — creating the first note adds an OOXML part, the problem `SetHeaderText`/`SetFooterText` already solved. `DocumentSnapshot` gains a `NoteParts` sibling field and `RestoreSnapshot` runs a `ReconcileNoteParts` **twin** of `ReconcileHeaderFooterParts` (separate method, same shape — not a rewrite of the existing one). Both branches are pinned: `DS328` (undo deletes the created part / redo re-creates it with its original relId) and `DS335` (undo of a *second* note leaves the part alone and rolls back only the definition).
- **`EditErrorCode.FootnoteRefNotSupported` is narrowed, not retired.** It stays — it is public surface clients switch on, and `MarkdownPayloadParserTests` asserts on it — but now means only "a `[^label]` in a markdown *payload* can't name a note the payload doesn't define". Its message points at the new ops. Deleting the enum member would be a breaking change for no gain.

### Ripple

Per `CLAUDE.md`: `DocxSessionOps` → WASM `DocxSessionBridge` + npm `insertFootnote`/`insertEndnote` → stdio `Dispatcher` (`insert_footnote`/`insert_endnote`) + `docx-scalpel` `insert_footnote`/`insert_endnote` → `docs/architecture/docx_mutation_api.md` → CHANGELOG.

## Verification

- **.NET: 3122 passed / 0 failed / 3 skipped** (full suite), including the new `DS320`–`DS338` and `DS340`–`DS345` batteries plus `GeneratedCssEscapingTests`. `DS334` runs `OpenXmlValidator` over a document carrying both an authored footnote and an authored endnote — zero errors.
- **Playwright: 277 passed / 0 failed** (full browser suite), including `npm/tests/docx-session-notes.spec.ts` (creation + round-trip, edit-and-delete through the existing ops, typed error envelope) and `npm/tests/editor-footnotes.spec.ts` (the editor renders markers + an editable notes section, edits persist, a citing paragraph keeps its citation).
- **Python: 53 passed / 0 failed**, including the new `python/tests/test_notes.py` (5 cases, end-to-end over the stdio host).
- **Release build clean** (warnings-as-errors): 0 errors. **WASM build succeeded**; `npx tsc --noEmit` clean.
- **Real-document smoke test** — the NVCA Model Certificate of Incorporation (234 paragraphs, 94 existing footnotes) driven through the browser `DocxEditor`: lossless round-trip, 0 new `OpenXmlValidator` errors (80 before / 80 after, all pre-existing), existing styles respected not duplicated, `settings.xml` untouched, and every LibreOffice marker paired with its own note. Details in the comment below.
- **LibreOffice round-trip** on a generated document — both notes render with their auto-numbers (`¹` footnote, `ⁱ` endnote) at exactly the requested offsets, with the separator rules and both note areas:

  ```
  The registrant¹ hereby amends tⁱhis registration statement.
  ⁱ  Cf. Item 512 of Regulation S-K.
  ¹  See Rule 473(a) under the Securities Act.
  ```

- Saved bytes verified free of internal `Unid` bookkeeping, and `word/footnotes.xml` / `word/settings.xml` match Word's shape byte-for-shape.

### Paginated footnote layout

Rendering notes exercised `pagination.ts`'s footnote engine on a dense real document for the first time and surfaced four silent defects, all fixed (`ece60fa`): notes dropped by a single-slot continuation that later notes overwrote; the stylesheet's `>` combinators XML-escaped so the rule keeping a note's number with its text was dropped by every browser; unsplittable notes re-wrapped inside themselves; and body text paintable underneath the note block. Reference document: **56 pages → 48, matching the LibreOffice render exactly**, zero overlaps, zero dropped notes. Detail in the third comment below.

### The editor renders what it can edit

Authoring a note is only half of it — the browser `DocxEditor` now **renders and edits** footnotes/endnotes (`2a56773`). Both render profiles emit notes; `AssignAnchorUnids` stamps the note parts so note paragraphs are ordinary `data-anchor` blocks (the whole ribbon works inside a note with no new command code); and the citation marker + backref are treated as inert converter chrome so the rendered display number can never be committed as literal text. Continuous mode renders a notes section; paginated mode activates `pagination.ts`'s existing footnote engine, placing notes at the bottom of the citing page. Also fixes Word-reserved `continuationNotice` notes projecting as user content. Full detail in the second comment below.

## Not yet (documented)

- Moving a citation without deleting and re-creating the note.
- Note numbering options (`w:numFmt`, restart-per-page/section, custom marks) — the `w:footnotePr` written on part creation declares only the separators; everything else is Word's default.
- Citing an existing note twice — each call creates a new definition.
- `TrackedChangeMode.RenderInline` does not wrap the citation in `w:ins`, consistent with every other insert op (`InsertParagraph`, `InsertTable`, `InsertHorizontalRule`, `SetHeaderText`); only `ReplaceText`/`DeleteBlock`/`DeleteRange` track.
- A session whose `ProjectionSettings.Scopes` exclude `Footnotes`/`Endnotes` writes the note correctly but returns no note anchors in `Created` — family behavior, same as `SetHeaderText` with `Headers` excluded.
- Endnotes render as a section after the paginated page stack rather than on their own final page — a layout imperfection; they are visible and editable.
- Paginated mode can still clip a few points of body text on a handful of pages (worst 16px, about one line) where the note reserve and the rendered note height disagree slightly; it clips cleanly rather than overlapping.


JSv4:
## Smoke test on a real document — found and fixed a shipping bug

Smoke-tested this against the [NVCA Model Certificate of Incorporation](https://nvca.org/wp-content/uploads/2025/10/NVCA-Model-COI-10-1-2025.docx) (234 paragraphs, **94 existing footnotes**, 18 header/footer stories) through the browser `DocxEditor` surface. That document was a good stress case precisely because it already has a dense footnote set — it exercises id allocation and part reuse rather than the create-from-nothing path the unit tests mostly cover.

### The bug

`max(id) + 1` is only correct when the new citation follows **every** existing one. Cite a note *earlier* in the document and it gets the highest id while being referenced first — which breaks an invariant every Word-authored document holds: **note ids ascend in reference order.** I verified that across the whole `TestFiles` corpus (including documents whose ids have gaps, e.g. 17/21/26) and the 94-footnote model certificate.

Renderers depend on it. LibreOffice numbers the body markers by citation position but pairs them against the **id-sorted** definition list, so a first-cited note holding the highest id renders the *wrong note text* — the marker reads "1" and points at somebody else's footnote. Nothing errors, and the file schema-validates. The document is simply, silently wrong.

Minimal repro (insert one note ahead of two existing ones):

```
            BEFORE                              AFTER
Alpha¹  →  ¹EXISTING NOTE ONE.        Alpha¹  →  ¹BRAND NEW NOTE INSERTED FIRST.
Beta.²  →  ²EXISTING NOTE TWO.        Beta.²  →  ²EXISTING NOTE ONE.
Gamma.³ →  ³BRAND NEW NOTE …          Gamma.³ →  ³EXISTING NOTE TWO.
```

On the real document, the marker after "…portfolio investment." resolved to *See Kumar v. Racing Corp.* instead of the note actually inserted there.

### The fix (`620796a`)

The citation is now inserted first carrying a placeholder id, so the note's id can be chosen from where that citation actually lands:

- **Follows every existing citation** (the common case) → `max(id)+1` as before, still scanning references as well as definitions so a document with gaps can't alias an existing note. Nothing else moves — zero churn.
- **Lands earlier** → the new note takes the smallest id cited after it, and every user note at or above that id shifts up by one — definitions *and* every reference to them in every part, since an endnote can be cited from inside a footnote. Notes cited earlier keep their ids. Word-reserved notes (`separator`/`continuationSeparator`/`continuationNotice`) are never renumbered; their ids sit below every user id, so shifting upward cannot collide.

Taking the *minimum* of the following ids rather than the first keeps this correct even on an input document that already violated the invariant.

New coverage: **DS336** (ordering), **DS337** (each shifted note keeps its own text), **DS338** (the shift follows a reference living in a peer part — an endnote cited from inside a footnote body — and flushes it).

**Worth stating plainly: this would have shipped.** It passed 16 tests, `OpenXmlValidator`, and a clean LibreOffice render — because every one of those inserted the citation *after* the existing notes.

### Verification after the fix

- Real document: body refs strictly ascending 2…96, zero duplicate ids, and LibreOffice now pairs every marker with its own note (`1 → See NVCA Model Documents…`, `2 → See Kumar…`).
- `OpenXmlValidator`: **80 errors before the edit, 80 after, 0 new** — the 80 are pre-existing quirks in the NVCA source.
- Styles were **respected, not duplicated** (the document already defines `FootnoteText`/`FootnoteReference`), and `settings.xml` was left untouched since the parts already existed.
- Editor round-trip lossless: 235 blocks in, edit, Save, reopen → 235 blocks and 96 footnote definitions.
- Full .NET suite **3114 passed / 0 failed / 3 skipped**.

### Two pre-existing findings, deliberately not addressed here

- **Footnotes are invisible in the editor surface.** `DocxSessionOps.RenderHtml` doesn't set `RenderFootnotesAndEndnotes`, so neither an authored note nor the document's own 94 render — confirmed by rendering the untouched original. You can author a footnote through the API but not see it in the editor. Outside this issue's scope; the obvious follow-up if the editor is meant to surface notes.
- **Editor saves are ~4.3× larger** (929KB vs 216KB on this document) because `DocxEditor.open` passes `persistAnchorIds:true` — an intentional tradeoff with an in-code comment explaining it. The note op itself adds **48 bytes** over a no-edit save.

JSv4:
## The editor now renders what it can edit (`2a56773`)

The previous comment listed "footnotes are invisible in the editor" as a pre-existing gap and moved on. That was the wrong call — an editor GUI that can author a footnote and can't show it isn't finished. Fixed.

### Three things were missing, each with its own failure mode

1. **The render profile had notes off** — hardcoded `false` in *both* the first-paint (`completeArgs` in `editor.ts`) and remount (`DocxSessionOps.RenderHtml`) paths. Both now enable them, each carrying a comment pointing at the other, because the remount output is required to match the first paint byte-for-byte. Unlike the header/footer bands this is **not opt-in**: notes are document content.

2. **Note paragraphs had no anchors**, so they would have rendered and stayed read-only. The full render assigned Unids only to `MainDocumentPart`. `HtmlConversionOps.AssignAnchorUnids` now covers the footnotes/endnotes parts too, and `FindByUnid` searches them so a single note can be re-rendered after an edit. Note paragraphs are therefore ordinary `data-anchor` blocks — **the entire existing ribbon works inside a note with no new command code.** Header/footer parts stay unstamped by design: paginated output clones one header node onto every page, so that anchor wouldn't be unique. Each note renders exactly once.

3. **The citation marker renders a visible number the session's run text doesn't contain** — the citation is a zero-width `w:footnoteReference`, the number is computed by the renderer from document order, and the backref (`↩`) is generated too. One `GENERATED_CHROME_SELECTOR` (shared with generated list markers) excludes all of it from the content-offset space, skips it when serializing a commit, and marks it non-editable. Missing any one fails differently: offsets drift (`OffsetOutOfRange`, silently dropped edits); the display number is **committed as literal text**, destroying the citation run; or the user deletes a marker and orphans its note.

### Both modes work — paginated better than expected

Continuous renders `section.footnotes` at the end of the body flow. Paginated turned out to be richer for free: `pagination.ts` already had a footnote engine (per-page distribution, splitting, continuations) that the render flag activates, so **notes land at the bottom of the page that cites them, above a separator rule**, continuing across pages when one doesn't fit. Endnotes render as a section after the page stack rather than on their own final page — a layout imperfection, not a correctness one.

A note split across pages puts a *different paragraph* in each half, so no two editable nodes ever share an anchor. (I briefly added a dedupe guard for this, then reverted it once I'd measured: it protected against nothing real and risked breaking the header/footer bands, which deliberately share unids across parts.)

### Rode-along fix: `continuationNotice` projected as a user note

`IsBoilerplateNote` enumerated only `separator`/`continuationSeparator`, so a `continuationNotice` — which real documents carry, including this one — reached the anchor index and the markdown projection as user content. Invisible until notes started rendering, at which point it showed as a stray empty footnote with no citation. Any **typed** note is now reserved: ECMA-376 §17.11.17 defines the type as `normal` | `separator` | `continuationSeparator` | `continuationNotice`, and only `normal` (which Word omits rather than writes) is user content. `DeleteBlock`'s reserved-note guard shares that one predicate so the two can't drift.

### Verified on the NVCA model certificate, through the browser editor

- **94 numbered citation markers** in the body, **110 editable note paragraphs**, markers and backrefs inert.
- Real keyboard typing **into a footnote** and **into a citing paragraph**: both persisted through Save → reopen, all 94 citations intact, no display number leaked into the text.
- The saved artifact: 94 body citations, 97 definitions, zero `Unid` leakage.
- **LibreOffice pairs every marker with its own note**, and renders both edits.
- **.NET 3120 / 0 / 3** — exactly +6 (DS340–DS345), so widening the reserved-note predicate moved nothing else. **Playwright 273 / 0.**

Three existing specs sampled "the first/last block" and now caught note content, since notes live in the body flow. Each was narrowed to body blocks rather than having its assertion weakened.

JSv4:
## Paginated mode fixed — now matches the LibreOffice reference page-for-page (`ece60fa`)

I sampled five random pages of the reference document, then swept all 56. Turning footnotes on had exercised `pagination.ts`'s footnote engine against a dense real document for the first time; it had per-page distribution, splitting and continuations already, but **four defects, every one of them silent.**

### 1. Notes were being dropped

An unfitted note was held in a **single** continuation slot, assigned inside a loop over the page's citation list. Two unfitted notes on one page → the second overwrote the first → that note rendered nowhere. **Four of 94 notes were missing from the output.** Deferred notes now queue and are *merged* into the next page's list; the three new-page branches previously clobbered what `finishPage` had just seeded.

### 2. The stylesheet's child combinators were XML-escaped

The generated CSS is the **value** of an `h:style` element, so serializing the XHTML turned `.footnote-content > p:first-of-type` into `… &gt; p…` — not a valid selector, so every browser dropped the rule. That is why each note printed its number alone on one line with the text below it. The rules now use descendant selectors, and `GeneratedCssEscapingTests` fails if *any* generated CSS reaches the browser escaped — the general form of the bug rather than the one instance.

### 3. Unsplittable notes were re-wrapped inside themselves

`splitFootnoteToFit` returned the whole `.footnote-item` as a "fitting element", which `addPageFootnotes` then spliced into a freshly built item — nesting a complete note inside another's content span, whose block-level div broke the same line. It now returns the note's content.

### 4. Body text could be painted underneath the notes

The note block is bottom-anchored and grows upward while the content area spanned the full text height, so only the layout loop's arithmetic kept them apart — ~134pt of superimposed, illegible glyphs on one page. The content area is now shrunk to what the notes actually leave, making the collision impossible **by construction**; worst case is a clean clip. (The shrink runs *after* the page box is attached — `getBoundingClientRect` on a detached node is all zeroes, so measuring during page construction silently did nothing.)

Separately, note heights are now measured with the `page-footnotes` class applied, so the reserve is computed in the same styling context (`font-size: .85em`, `line-height: 1.4`) the notes render in. That alone took worst-case body clipping from 144px to 16px.

### Result on the reference document

| | before | after | LibreOffice |
|---|---|---|---|
| pages | 56 | **48** | **48** |
| overlapping pages | 1 | **0** | — |
| dropped notes | 4 | **0** | — |
| worst body clip | 144px | **16px** | — |
| page occupancy | 91% | 94% | — |

### Two corrections to my earlier report

**The "43% empty pages" defect did not exist.** I measured body extent against the full content box and counted the note band — which legitimately occupies the rest of the page — as wasted space. True occupancy was 91%, not 57%. The comparison I offered as proof ("93% with notes off, 57% with notes on") was the tell and I read it backwards: that delta is arithmetic, not a regression.

**I reverted one of my own changes.** A `remainingHeight` "consistency fix" in the footnote-expansion branch. The two branches do genuinely disagree about whether that variable has the footnote reserve folded in, but measured against the real document the change *tripled* note loss (4 → 12), and no failing test justified it. Dropped.

### Verification

- **.NET 3122 / 0 / 3**, **Playwright 277 / 0.**
- The new layout battery (`npm/tests/pagination-footnote-layout.spec.ts`) pins: no body/note overlap, no page left idle while blocks are pending, no cited note dropped, no nested notes. Its content-loss guard is **verified discriminating** — it fails with the deferral fix reverted and passes with it.

PR #337, opened 2026-08-03
feat(docxdiff): Word Compare "Comparison settings" parity (white space, textboxes, fields, tables, character granularity)

rey-ms:
Adds the missing **Word Compare "Comparison settings"** options to the `DocxDiff` engine, plus two rendering-correctness fixes found while building them. Every option defaults to Word's own default, so **existing output is byte-identical unless a caller opts in**.

`docs/architecture/wml_comparer_gaps.md` gained a table mapping Word's Compare dialog onto our settings; this series fills in five of its rows.

### Added

| Word Compare option | Setting | Default |
|---|---|---|
| White space | `WmlComparerSettings.CompareWhitespace` / `DocxDiffSettings.CompareWhitespace` | `true` |
| Textboxes | `DocxDiffSettings.CompareTextboxes` | `true` |
| Fields | `DocxDiffSettings.CompareFields` | `true` |
| Tables | `DocxDiffSettings.CompareTables` | `true` |
| Show changes at (Character/Word level) | `DocxDiffSettings.ChangeGranularity` | `Word` |

- **White space** — off canonicalizes whitespace in *both inputs* before the diff (shared `Internal/WhitespaceCanonicalizer.cs`, used by both engines), so whitespace-only differences produce no revisions. A tab is never equated with a space.
- **Textboxes** — a *granularity* switch, not suppression: off, a changed box is deleted-and-reinserted wholesale by its host paragraph instead of inner-diffed, so the change stays in the markup.
- **Fields** — governs whether field *codes* (instruction text, simple-vs-complex form, `w:fldChar`/`w:fldData` scaffolding) are compared. Field *results* are ordinary text either way. On, a code change is one reversible whole-paragraph `w:del`/`w:ins` pair carrying `w:delInstrText`.
- **Tables** — governs table content/structure (rows, cells, per-cell text). A whole table added, removed or moved is a block-level operation and is still reported with the box off; a shell-only change (column widths) routes to `FormatOnlyBlock` so table *formatting* stays on its own axis (`TrackBlockFormatChanges`).
- **Show changes at** — `Character` narrows a change to the differing characters (`colour` → `color` renders retained `colo` + deleted `u` + retained `r`) in both the markup and the revision list. Alignment stays word-grained; the edit script is identical under both values.

### Fixed

- **A changed shape or textbox no longer produces a schema-invalid document.** A changed drawing renders as a deleted copy plus an inserted copy, both carrying the source's ids, so two revisions of one document produced duplicates (`Sem_UniqueAttributeValue`) across `wp:docPr/@id`, the VML shape-id space, `v:shapetype/@id` and `o:spid`. A new `NormalizeDrawingIds` pass re-ids the non-surviving copies, choosing the keeper by revision side so the id lands on the copy that survives accept.
- **A change inside a textbox is now tracked inside the box.** Previously the change was *detected* but not rendered, so two documents differing only in one textbox's text compared as unchanged. The renderer now emits per-box interior revisions, falling back to the conservative whole-block replacement when a box cannot be sliced.

### Round-trip contract

`accept ≡ right` and `reject ≡ left` hold for every option at its default. Where an option is turned **off**, the uncompared difference is by definition not reversible — one side's content rides through untracked, and which side depends on the scope (body renders from the right; a header/footer story or note definition keeps its left part). This is stated per option in the XML docs and in `docs/architecture/ir_diff_engine.md`.

### Commits

The series has real internal dependencies and is ordered accordingly:

1. `feat(comparer)` White space (`CompareWhitespace`) — introduces the shared canonicalizer and the parity table
2. `feat(docxdiff)` White space on the `DocxDiff` engine
3. `feat(docxdiff)` Textboxes (`CompareTextboxes`)
4. `fix(docxdiff)` unique drawing ids in compare output
5. `feat(docxdiff)` track a textbox change inside the box — builds on 4
6. `feat(docxdiff)` Fields (`CompareFields`)
7. `feat(docxdiff)` Show changes at (`ChangeGranularity`) — builds on 4
8. `feat(docxdiff)` Tables (`CompareTables`)

### Verification

- Full suite: **3475 passed**. The 7 failures are all pre-existing on `main` at `44dac09` (`DS310`–`DS312`, `MD018`, `MD030`, `IrCorpusTests DB007-Notes.docx`, `MCP137`) — confirmed by running them on an unmodified checkout.
- `dotnet build -c Release Docxodus.sln` (warnings-as-errors): 0 errors.
- `./scripts/build-wasm.sh`: exit 0. `npx tsc --noEmit` in `npm/`: clean.
- Each setting is covered by a dedicated test class, and each was **mutation-tested** — the gate was inverted/removed and the suite confirmed to fail on exactly the intended tests. Where a mutant survived, a discriminating fixture was added.
- Client surfaces updated in step: `DocxDiffOps` wire keys, `npm/src/types.ts`, `python/src/docx_scalpel/`.


JSv4:
These gaps should largely be closed in `DocxDiff`. Apologies for the lack of clarity on `WmlComparer` being in maintenance mode. Docs corrected in #364. Please add all new changes to `DocxDiff`.

PR #356, opened 2026-08-06
Add a CompareTables option to WmlComparer

rey-ms:
Adds `WmlComparerSettings.CompareTables`, mirroring the **Tables** checkbox in Word's Compare Documents dialog.

Default is `true`, so existing callers are unaffected and output is byte-identical to before.

## Behaviour when off

Body-level tables take no part in the comparison: an edited, added or removed table produces no revision, and the result carries the original document's tables unmarked. As with any ignored scope, accept and reject therefore no longer reproduce the revised document's tables.

## How

Each of the original document's body-level tables is swapped for a marker paragraph before hashing, so the marker rides through the LCS in the table's place and the real table is laid back over it afterwards. The revised document's tables are dropped: an ignored table is never carried over, and marking both sides would let a table count difference perturb how the surrounding prose matches.

The marker token is generated per comparison, so document text cannot impersonate it. Restoration happens before footnote rectification and the docPr/shape id fixups, so a note cited only inside an ignored table still resolves and ids stay unique. The atom/LCS core is untouched.

## Scope

Tables that are direct children of `w:body`. A table nested in a content control or a textbox is still compared.

`DocxCompare` drops the option on the `DocxDiff` branch (that engine has no ignore-tables mode), and it is not surfaced on WASM/npm — matching the sibling `DetectFormatChanges`.

## Tests

`Docxodus.Tests/WmlComparerCompareTablesTests.cs`, 15 cases: block-position preservation, several tables, differing non-zero table counts on the two sides, a table removed or moved ahead of other tables, a right-only table, a table-only document, marker-lookalike prose, `Consolidate`, and a footnote cited only inside an ignored table. Every case asserts a schema-valid package with no marker left behind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- **`WmlComparerSettings.CompareTables` — Word's "Tables" compare checkbox (default `true`, so
+  existing callers are unaffected).** When false, body-level tables take no part in the comparison:
+  an edited, added or removed table produces no revision, and the result carries the original
+  document's tables unmarked — so, as with any ignored scope, accept and reject no longer reproduce
+  the revised document's tables. Each of the original's body-level tables is swapped for a marker
+  paragraph before hashing so it lands back in place afterwards; the atom/LCS core is untouched, so
+  nothing changes when the option is left on. v1 scope: tables that are direct children of
+  `w:body` — one nested in a content control or a textbox is still compared. Coverage:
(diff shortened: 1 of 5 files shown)


JSv4:
Hey @rey-ms, I'm not planning to continue to maintain WmlComparer. A reason why you'd prefer that over DocxDiff?

rey-ms:
@JSv4 From what I understand, WmlComparer is the "stable" engine, that's why I'm using it.

JSv4:

BTW, this functionality should be in DocxDiff but please feel free to open an issue and/or cut a PR if it doesn't work the way you want it to.

PR #357, opened 2026-08-06
Fix a crash comparing a document whose body has no paragraphs

rey-ms:
`WmlComparer.Compare` and `GetRevisions` throw `NullReferenceException` when one side's `w:body` contains no `w:p` elements — only a `w:sectPr`.

```
System.NullReferenceException
   at Docxodus.WmlComparer.MoveLastSectPrIntoLastParagraph(XElement contentParent)
   at Docxodus.WmlComparer.CreateComparisonUnitAtomList(...)
   at Docxodus.WmlComparer.ProduceDocumentWithTrackedRevisions(...)
   at Docxodus.WmlComparer.Compare(...)
```

## Reproduction

Take any `.docx`, remove every `<w:p>…</w:p>` from `word/document.xml` leaving the trailing `<w:sectPr>`, and compare it against the original. Direction does not matter — an empty body as either side throws. Word's UI always keeps at least one paragraph, so this is only reachable with programmatically generated input.

## Cause and fix

`MoveLastSectPrIntoLastParagraph` looks for the paragraph to move the trailing `w:sectPr` into and dereferences the result without checking it found one. The fix returns early, leaving the section as a body child — which is where the produced document puts it back regardless, so nothing is duplicated, lost or misplaced.

The same three lines also called `lastParagraph.AddFirst(W.pPr)`, adding the XName rather than the `pPr` element just created. That left the element detached, so the `w:sectPr` appended to it was dropped and the name was added as a text node. That branch is reached routinely — a last paragraph with no `w:pPr` is ordinary — but it was output-neutral: the stray text node is never read (the atom walk enumerates elements only) and the section is re-added unconditionally from the original document at the end of `ProduceDocumentWithTrackedRevisions`.

## Tests

`Docxodus.Tests/WmlComparerEmptyBodyTests.cs`: the paragraph-less body as either side and as both, the helper's created-`pPr` behaviour asserted directly, and a body whose only block content is a table (the one shape where the early return deliberately does not fire, and the section moves into a cell paragraph). Each test was confirmed to fail without its corresponding fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- **`WmlComparer.Compare` and `GetRevisions` no longer throw `NullReferenceException` when a document
+  body contains no paragraphs.** `MoveLastSectPrIntoLastParagraph` looked for the paragraph to move the
+  trailing `w:sectPr` into and dereferenced the result without checking it found one, so a body holding
+  only a `w:sectPr` — which Word's UI cannot produce but a generator can — crashed as either side of the
+  comparison. The section is now left as a body child, which is where the produced document puts it back
+  regardless. The same three lines added the `w:pPr` XName rather than the element just created, leaving
+  it detached, so the section appended to it was dropped and the name became a text node; that branch is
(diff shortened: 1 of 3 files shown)


JSv4:

PR #361, opened 2026-08-07
Fix an infinite loop comparing a table against a text box

rey-ms:
`WmlComparer.Compare` never returns when one document contains a table and the other a text box in the same position. It is not a deadlock and not an expensive search: one core runs flat out, memory stays completely flat, and the call simply never ends. Callers see it only as a timeout.

### Cause

The structural walk in `DoLcsAlgorithm` groups each side into adjacent runs keyed `Word` / `Row` / `Textbox`, then advances `iLeft` / `iRight` until both are exhausted. Every branch of the chain required one side to be a `Word`:

| left | right | branch |
|---|---|---|
| same | same | advance both |
| `Word` | `Row` | advance right |
| `Row` | `Word` | advance left |
| `Word` | non-`Word` | advance left |
| non-`Word` | `Word` | advance right |
| **`Row`** | **`Textbox`** | **none — neither advances** |

Two *different non-word* kinds match nothing, so neither counter moves and the `while (true)` spins, allocating nothing.

### Fix

A final `else` retires the left group as a deletion and advances `iLeft`. That restores the invariant which makes the walk terminate at all — every branch now advances at least one counter — while leaving the right group free to pair with a later group of its own kind (retiring both sides would render everything as a deletion plus an insertion).

The sibling table-vs-paragraph walk later in the same method was checked and is not affected: its key space is only `Table`/`Para` and all four combinations already advance.

### Evidence

Found by fuzzing a corpus of 1,925 documents, comparing 400 random pairs drawn only from files that self-compare cleanly. 7 pairs (~1.8%) hung past 600 s, every one of them mixing tables with text boxes. All 7 now complete in milliseconds.

Each pair was independently reproduced with a minimal synthetic fixture before the fix was written; the hot frame was confirmed by stack sampling rather than inspection.

### Tests

`WmlComparerRowVersusTextboxTests` (RT001–RT003) — a row against a text box, the mirror case, and the mismatch reached with differing prose on both sides. Because a regression here hangs rather than fails, each test asserts a deadline on a worker thread, and asserts that neither side's content was dropped.

All three were verified to hang with the fix reverted and to pass with it, so they are not vacuous. Full suite is unchanged otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,18 @@ All notable changes to this project will be documented in this file.
   the CRUD rules are documented in `docs/architecture/docx_mutation_api.md`.
 
 ### Fixed
+- **`WmlComparer.Compare` no longer hangs forever comparing a document containing a table against
+  one containing a text box.** The structural walk in `DoLcsAlgorithm` groups each side into adjacent
+  runs of Word / Row / Textbox and advances a counter per side, but every branch required one side to
+  be a `Word`. Two *different non-word* kinds facing each other — a row against a text box — therefore
+  matched no branch and advanced neither counter: a loop that allocated nothing and never terminated,
+  pegging one core with flat memory until the caller gave up. The left group of such a pair is now
+  retired as a deletion, restoring the invariant that makes the walk terminate at all (every branch
+  advances at least one counter) while leaving the right group free to pair with a later group of its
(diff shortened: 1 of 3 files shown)


JSv4:
We're not maintaining WmlComparer further, @rey-ms. Thank you for the contribution though.

JSv4:

PR #362, opened 2026-08-07
Add paragraph-level format change tracking to WmlComparer

rey-ms:
`WmlComparerSettings.DetectParagraphFormatChanges` (default **false**) tracks alignment, indentation, spacing, paragraph style and list membership as native `w:pPrChange` markup, mirroring the existing run-level pass that emits `w:rPrChange`.

### Why

A paragraph-format-only change is currently *silently lossy*, not merely untracked.

A paragraph mark hashes as its element name plus its text value, and every `w:pPr` child is attribute-only. So a paragraph whose formatting alone changed hashes identically on both sides, correlates as `Equal`, and is rebuilt from the **right** document. The revised formatting is therefore already applied to the result and the original's discarded, with nothing reported — leaving `reject` unable to restore it.

### How

`ReduceToParagraphPropertiesChange` produces both the comparison key and the archived copy, so what is compared is exactly what is stored. It excludes what a `w:pPrChange` may not carry — its inner `w:pPr` is a `CT_PPrBase`, so the mark's own `w:rPr` and an inline `w:sectPr` are dropped — and strips revision-save ids and internal bookkeeping at every level, so they neither trip a false positive nor leak into the archive.

Accept and reject needed no new code: `RevisionProcessor` already handles `w:pPrChange`.

Lists ride the same path for free (`w:numPr` is an ordinary `w:pPr` child). Renumbering caused by editing `numbering.xml` is not a `w:pPr` change and is not tracked — Word does not track that either.

### A crash fixed in passing

Adding the emission branch also closed a crash reachable **without** this option, under default settings.

Run-level detection can stamp `FormatChanged` on a paragraph mark inside a text box: a paragraph there hangs off the run carrying the drawing, so unlike an ordinary paragraph mark it has a run ancestor, whose properties the run pass reads. The paragraph-properties transform had no branch for that status and threw `Internal error - unknown status: FormatChanged` on text-box documents.

### Two pre-existing defects on the default-on run path

Both affect `w:rPrChange` today and are fixed here because the paragraph path shares the code:

- Archived properties round-trip through an internal attribute. Without an explicit namespace declaration XLinq wrote them as a default-namespace `<rPr xmlns="…">` that re-declared the namespace on every child — and that survived into the produced document. Only visible when the old `w:rPr` was non-empty.
- A property with no `w:val` is compared by its serialized form when reporting changed property names, so an untouched `w:rFonts` counted as changed whenever the two copies carried different internal ids.

### Scope and limits

- **Default off**, because enabling it changes the output of comparisons that produce no revision today.
- `GetRevisions()` reports these as `FormatChanged` alongside run changes, carrying the paragraph's text. The revision type enum is unchanged. Reporting is deliberately *not* gated by the setting — a document that already contains a `w:pPrChange` is reported either way, exactly as `w:rPrChange` already was; that is what keeps a document produced with the option from going silent when read back with defaults.
- Body scope only: no formatting change inside a footnote or endnote is detected. This is pre-existing and shared exactly with the run-level pass — a *text* change in the same note is detected, so notes are compared and it is the format passes that miss them. Pinned by `PF022`, which fails if that ever changes.
- The paragraph mark's own run properties (`w:pPr/w:rPr`) and section changes remain out of scope.
- CLI: `redline --detect-paragraph-format-changes`.

### Tests

`WmlComparerParagraphFormatTests` — 25 tests covering detection (alignment, indent/spacing, style, list add/remove/level/format), accept and reject round-tripping the properties, `GetRevisions` reporting, independence from `DetectFormatChanges`, and the schema validity of every output.

The false-positive guards carry more weight than the positive cases and are the reason several of them exist: identical documents, revision-save-id-only differences, the same properties in a different source order (at both levels), and a section-only difference must each produce **zero** revisions.

One run-level test was added to `WmlComparerFormatChangeTests` for the changed-property-name fix, since that path is on by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,39 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- **`WmlComparerSettings.DetectParagraphFormatChanges` — paragraph-level format tracking as native
+  `w:pPrChange` markup (default `false`, so existing callers are unaffected).** Covers alignment,
+  indentation, spacing, paragraph style and list membership (`w:numPr`), and is independent of
+  `DetectFormatChanges`, which stays run-level (`w:rPrChange`) and stays on.
+  What it fixes is a silent loss rather than a missing feature: a paragraph mark hashes as its element
+  name plus its text value, and every `w:pPr` child is attribute-only, so a paragraph whose formatting
+  alone changed hashes identically on both sides, correlates as Equal, and is rebuilt from the RIGHT
+  document. The revised formatting was therefore already being applied to the result, the original's
+  discarded, and nothing reported — leaving reject unable to restore it. Detection mirrors the
+  run-level pass; `ReduceToParagraphPropertiesChange` produces both the comparison key and the archived
+  copy, so what is compared is exactly what is stored. It drops what a `w:pPrChange` may not carry —
+  its inner `w:pPr` is a CT_PPrBase, so the mark's own `w:rPr` and an inline `w:sectPr` are excluded —
+  and strips revision-save ids and internal bookkeeping at every level so they neither trip a false
+  positive nor leak into the archive. Accept/reject needed no new code: `RevisionProcessor` already
+  handles `w:pPrChange`. Adding the emission branch also closed a crash reachable without this option:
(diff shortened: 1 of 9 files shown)


JSv4:

PR #366, opened 2026-08-08
fix(editor): show what is being dragged and where it will drop

JSv4:
## The bug

The block drag handle floats in the page **margin**, 32 px left of the block's text column. So the natural gesture — press it and pull straight down — keeps the pointer in the gutter, where it never crosses a paragraph box.

Drop position was resolved by asking which element the pointer was over. For that gesture the answer was "nothing", for the whole duration: no drop line, and a release that silently did nothing. Only a drag steered into the text column worked — which is why drops appeared to succeed with no preceding feedback at all.

Measured on the pre-change build, dragging straight down the gutter:

```
GUTTER indicator shown:  []                      (never)
GUTTER result order:     ["One","Two","Three"]   (drop was a no-op)
```

Same gesture after:

```
GUTTER indicator shown:  [block, block, …]
GUTTER result order:     ["Two","Three","One"]
```

## The fix

**Drop position is geometry, not hit testing.** One drop target on the document flow; `resolveDropAt` picks the nearest block by vertical distance, and the half the pointer is in picks the side — snapped to the other side when only that one is legal. A block the source cannot legally reach still offers no drop and draws nothing: the line never snaps to a distant legal boundary the user did not point at, so the existing section-break partitioning holds unchanged.

Blocks are measured once at drag start. Nothing reflows mid-drag, and scrolling — including drag autoscroll — moves the flow rigidly, which one subtraction corrects.

**Three signals, each answering a different question:**

| Signal | Answers |
|---|---|
| Source block dims to 38 % | *what* is moving |
| Preview chip naming the block follows the cursor | *what* is under the cursor (the browser otherwise ghosts the 26 px grip) |
| 2 px accent line with a leading dot | *where* it will land |

The line is positioned by `transform`, so tracking the pointer costs no layout, and gets a one-shot 110 ms fade per appearance rather than a transition that would smear while tracking.

## Performance

Measured by `editor-move-latency-bench.spec.ts` on `NVCA-Model-COI.docx` (234 body blocks, 392 bookmarks, 3 section breaks):

| | Before | After |
|---|---|---|
| Per-drag-start block work | 1.5 ms (`refreshBlockDropTargets`) | 1.3 ms (`captureDropZones`) |
| Per `dragover` | — | 0.0 ms (`resolveDropAt`, new budget 5 ms) |
| Registered drop listeners on the flow | 234, re-registered at every drag start | 1, registered once |

A wash on the per-gesture cost; the win is correctness plus 233 fewer listeners. The bench's `refreshBlockDropTargets` row is replaced by the two rows above.

## Tests

`editor-block-drag.spec.ts` gains a regression test that drags down the gutter **without ever entering the text column** (asserted), sampling per pointer step rather than only at the end: the line appears as soon as the pointer clears the source block and then tracks continuously, at the block's width, to more than one position; the source stays dimmed; the preview names the block; and what the line predicted is where the block lands. Cleanup is asserted too — nothing left dimmed, line hidden.

- `editor-block-drag.spec.ts` — 6/6, run 5× for flake
- Full chromium suite — 327 passed. The 17 `worker.spec.ts` / `worker-prepare.spec.ts` failures in this environment reproduce on a clean tree with these changes stashed, so they are pre-existing and unrelated.

## Docs

- `CHANGELOG.md` under `[Unreleased] / Fixed`
- `docs/architecture/editor_ui_surface.md` — new "Drag feedback, and why drop position is geometry" subsection; corrects the now-stale claim that drop targets are registered per block at drag start

No public API change: no new setting, no WASM/npm surface change, `blockDrag` is unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_014dVpwDeHP8cygerAkvBA27)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ All notable changes to this project will be documented in this file.
     block follows the cursor (replacing the browser's ghost of the 26 px grip), and a 2 px accent
     line with a leading dot marks the insertion point, positioned by `transform` so tracking the
     pointer costs no layout.
+  - The line is drawn in the MIDDLE of the gap between two blocks, not on the target's border-box
+    edge. A paragraph's `w:spacing` becomes a CSS margin, which sits outside the box, so an
+    edge-drawn line underlined the block's last line instead of reading as a gap between blocks.
 - **The editor's continuous view lays the document out at its own page width, and zooms to fit a
   narrow window instead of letting content run off the page.** Enlarging a block's font size on a
   phone pushed text and tables past the sheet, where the window clipped them. Three layers were
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Fixed
+- **Dragging a block in the editor now shows what is moving and where it will land.** The drag
+  handle floats in the page margin, so the natural gesture — press it and pull straight down —
+  keeps the pointer in the gutter, where it never crosses a paragraph box. Because the drop
+  target was resolved by asking which element the pointer was over, that gesture found no target
+  for its whole duration: no drop line, and a release that silently did nothing. Only a drag
+  steered into the text column worked, which is why drops appeared to succeed with no preceding
+  feedback at all.
+  - Drop position is now resolved from the pointer's **vertical position** against the blocks
+    (`resolveDropAt`), not from element hit testing, so the gutter counts as being over the
+    document. Blocks are measured once per drag; scrolling is corrected by one subtraction.
+    Cost is 0.0 ms per `dragover` on a 234-block charter, and the flow no longer carries one
+    registered drop listener per block.
+  - A block the source cannot legally reach (across a section break, say) still offers no drop
+    and draws nothing — the line never snaps to a distant legal boundary the user did not point at.
+  - Three signals now run for the whole gesture: the source block dims, a preview chip naming the
+    block follows the cursor (replacing the browser's ghost of the 26 px grip), and a 2 px accent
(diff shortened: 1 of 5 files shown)


JSv4:
CI status on `14b428b`:

- **Playwright Tests — ✅ passed.** This is the suite that covers the change, including the new gutter-drag regression test.
- **Claude Code Review — ✅ passed.**
- **CI (`build-and-test`) — ❌ failed on one .NET test**, which is a wall-clock flake unrelated to this diff:

```
Docxodus.Tests.Ir.Diff.IrAlignerAdversarialTests.Scale_guard_500_vs_2000_wall_ratio_within_12x
Align wall-time ratio 14.22x for 4x input exceeds the 12x anti-O(n²) guard
(500=3.23ms, 2000=45.93ms)
```

Total: 3410 tests, 1 failed.

Why it isn't mine: this PR changes **only TypeScript** (`npm/src/editor.ts` and two specs) plus CHANGELOG/docs — no C# is touched, so it cannot affect the IR aligner. `main` is green at `93fd230`, so this isn't a base-branch breakage either. Re-running that test on this branch locally passes:

```
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1
```

The guard compares wall time at n=500 (≈3 ms) against n=2000, and a ~3 ms denominator is small enough that runner scheduling noise dominates the ratio — the test's own docstring records a previous flake of this kind ("observed 8.72× flake in a full-suite run vs 4.7× in isolation"). Making that guard robust is a separate change and doesn't belong in this diff.

I don't have permission to re-run the failed job (`rerun-failed-jobs` returns 403), so it needs a re-run from someone who does. Happy to push a robustness fix for the guard in its own PR if that's wanted.

---

PR #460, opened 2026-08-14
Remove the vendored CPOL-licensed PEG parser and Excel formula path

JSv4:
> ⚠️ **Breaking change.** Removes public API. Needs a major version bump at release time.

## The problem

`Docxodus/PegBase.cs` line 4:

```
/*Author:Martin.Holzherr;Date:20080922;Context:"PEG Support for C#";Licence:CPOL
```

…sitting directly beneath lines 1–2, which this repository's StyleCop header template stamped on:

```csharp
// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```

Two incompatible license claims in a single file. CPOL (Code Project Open License) is not OSI-approved and is generally treated as incompatible with MIT redistribution — yet this file ships inside a package declaring `<PackageLicenseExpression>MIT</PackageLicenseExpression>`, and neither `LICENSE` nor `README.md` discloses it.

## Why delete rather than disclose

Everything the parser supported turned out to be unreachable:

| File | Lines | Status |
|------|-------|--------|
| `PegBase.cs` | 2109 | Third-party, CPOL, 2008 |
| `ExcelFormula.cs` | 833 | **Machine-generated** in 2012 from an `ExcelFormula.txt` grammar that **is not in this repository** — it could not be regenerated or maintained |
| `SSFormula.cs` | 105 | The only PegBase consumer |

Their only downstream consumers were `WorksheetAccessor.FormulaReplaceSheetName` and `WorksheetAccessor.CopyCellRange`. I checked for callers across the library, tests, all three CLI tools, the WASM bridge, the npm package and the Python client:

```
$ grep -rn "FormulaReplaceSheetName\|CopyCellRange" --include=*.cs --include=*.ts --include=*.py .
(no results outside WorksheetAccessor.cs itself)
```

`WorksheetAccessor` has **no test coverage at all** — it is one of three modules (with `ExcelFormula` and `TextReplacer`) with zero references anywhere in `Docxodus.Tests/`.

Keeping 3047 lines of unmaintainable generated code under a conflicting license, to support two methods nobody calls, is not a trade worth making.

## What disappears

[...]

PR #461, opened 2026-08-14
Bound the undo ring by memory, not just step count

JSv4:
## The problem, measured

`UndoDepth` never bounded memory. Each undo step is a deep clone of every snapshot-scoped part, so **one step costs whatever the document costs** — a fixed step count over a large document is an unbounded amount of RAM.

Measured on `TestFiles/NVCA-Model-COI.docx`, **144 KB on disk**:

| | |
|---|---|
| One snapshot | **≈7.5 MB** retained |
| Old default (50 steps) | **≈375 MB** of live DOM — for a 144 KB file, with nothing bounding it |
| New default (20 steps) | ≈150 MB worst case, and the budget binds first |

Fine on a server. A latent OOM in a browser WASM heap — which is where `DocxEditor` runs, and no client overrode the default.

## The fix

`DocxSessionSettings.UndoMemoryBudgetBytes`, default **128 MiB**. The ring evicts oldest-first until under **both** bounds, with two deliberate policies:

- **Redo is surrendered before undo.** Redo is speculative — it only exists between an `Undo()` and the next mutation. Undo history is the user's own.
- **The last undo step is never evicted.** A single snapshot larger than the whole budget would otherwise make undo silently unavailable on exactly the large documents where a mistaken edit is most expensive to lose.

`UndoDepth` drops 50 → 20 so the two bounds roughly agree on a typical document (~17 steps at 128 MiB) and the budget takes over as documents grow. Set `UndoMemoryBudgetBytes = 0` for the previous depth-only behaviour — which also skips measurement entirely (`UR005` asserts the cost selector never runs).

## Why not measure serialized size

A live LINQ-to-XML tree runs several times its markup size — every element, attribute and text node is a heap object with headers and parent/sibling links. Budgeting against serialized bytes would under-count by roughly an order of magnitude, which is precisely the error that matters for a memory bound. `XmlMemoryEstimator` models retained heap instead, with constants documented as a model rather than a measurement.

### Estimator cost

It runs on every mutation, beside the clone it measures, so its cost is load-bearing:

| Implementation | Cost, relative to the clone |
|---|---|
| `DescendantNodes()` + `Attributes()` | **167%** — more than doubled snapshot cost |
| Raw `FirstNode`/`NextNode` walk (shipped) | **48%** |

The LINQ helpers allocate an iterator per element and run a state machine per node. The manual walk is allocation-free and uses an explicit stack rather than recursion so a pathologically nested document can't overflow.

On the 144 KB document: `TakeSnapshot` 4.87 ms → estimator adds 2.36 ms. Against the editor's measured 200–360 ms structural ops that is noise; for callers on the fast path who disagree, the opt-out is one setting.

## Observability

The bound is silent otherwise, so it is now reportable:

- `UndoCount` / `RedoCount`
- `UndoMemoryBytes` — approximate retained heap (always 0 when the budget is disabled, since nothing is measured)
- `UndoHistoryTrimmedForMemory` — **sticky**, and specifically means *the budget, not the depth* discarded history

That last one is the difference between an editor explaining "history was trimmed to stay within memory" and an editor that just looks broken when undo stops short of the configured depth.

## Rode-along: duplicated wire defaults

`DocxSessionJson.ParseSettings` and the MCP dispatcher **each hardcoded `undoDepth = 50`** independently of the .NET default. Changing the default in `DocxSessionSettings` alone would have left both surfaces silently on the old value. Both now read the defaults from a `DocxSessionSettings` instance, and `UR023` asserts the wire defaults track the .NET ones.

## Client ripple

Per `CLAUDE.md`'s checklist: `npm/src/types.ts`, `python/src/docx_scalpel/types.py` (field + `to_wire`), `tools/mcp-server/ToolCatalog.cs` + `Dispatcher.cs`.

## Verification

```
Release build:  0 errors
Full suite:     3447 passed, 0 failed, 3 skipped  (was 3435 — +12 new)
Python client:  types.py parses, wire key + field present
TypeScript:     interface-only addition; tsc here fails on a missing
                @types/node (deps not installed in this environment),
                unrelated to the change
```

12 new tests (`UR001`–`UR023`): depth-vs-budget precedence, eviction order and survivor identity, the never-evict-last guarantee, opt-out skipping measurement, byte accounting across pop/clear, estimator scaling and its floor above serialized length, and wire defaults tracking .NET defaults.

## What this does *not* do

This **bounds** the cost; it does not reduce it. 7.5 MB per snapshot for a 144 KB document is the real inefficiency, and the fix for that is delta snapshots rather than whole-part clones — a much larger change, and a separate one. Worth noting that [#459](https://github.com/JSv4/Docxodus/pull/459) adds `settings.xml` + `styles.xml` to the snapshot scope, which nudges per-snapshot cost up; these two land cleanly together but the combined effect on snapshot size is worth a look.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mh67RmyGM9HFqyTeukwWop)_


JSv4:
Playwright is red here for the same base-branch reason as [#460](https://github.com/JSv4/Docxodus/pull/460#issuecomment-5289122010) — full detail in that comment, not repeated here.

```
[chromium] › tests/social-demo.spec.ts:363:3
  The landing page on a phone › the floating controls steer the game and keep the screen clear
1 failed · 9 skipped · 483 passed
```

Same failure on `main` at c391985 (#433, the commit that added the test), with identical counts. This branch merged `main` twice — once after #462, once after #459 — and inherited it.

This PR is .NET-only (`DocxSession`, `UndoRing`, the new `XmlMemoryEstimator`) plus type declarations in `npm/src/types.ts` and `python/src/docx_scalpel/types.py`. It touches no demo or arcade code. Locally on the merged branch:

```
Release build (library + tests + mcp-server):  0 errors
Full suite:  3456 passed, 0 failed, 3 skipped
```

`build-npm` and both `python-tests` jobs are green here, which is what actually covers the client changes.

Ready to merge from my side once the base recovers.

---

Issue #490, opened 2026-08-15
A batch step that matches nothing fails as internal_error; the direct call silently no-ops


JSv4:
Found by the epic #435 acceptance smoke against `main` (`400af83`). Reproduced, not code-read.

`replace_text_range` with a `find` string that is genuinely absent from the target anchor behaves two different — and both unhelpful — ways depending on which surface calls it.

### Direct call: silent no-op

```json
{ "sessionId": "…", "action": "replace_text_range",
  "anchorId": "p:body:…", "find": "ZZZ-not-present-ZZZ", "replace": "x" }
```

returns

```json
[]
```

No error, no count, no signal. An agent cannot distinguish "I replaced nothing because nothing matched" from "I replaced something", because the success shape for one replacement is a one-element array and the success shape for none is an empty one. The natural agent behaviour — assume it worked — is wrong, and nothing in the response says so.

### Same op inside a batch: `internal_error`

```json
{ "sessionId": "…", "mode": "atomic", "steps": [
  { "tool": "docxodus_edit", "args": { "action": "replace_text_range",
      "anchorId": "p:body:…", "find": "ZZZ-not-present-ZZZ", "replace": "x" } } ] }
```

returns `status: "failed"`, `rolledBack: true`, and

```json
{ "code": "internal_error", "message": "batch mutation returned no valid edit results" }
```

The batch layer maps "the op produced zero edit results" onto `internal_error`. That is the wrong error class: matching nothing is an ordinary, fully expected outcome of a search-and-replace, not an internal fault. It also makes the message leak an implementation detail ("no valid edit results") in place of the thing the caller needs to know — which step, which anchor, and that the needle was not found.

### Why it matters

`internal_error` is the one code an agent cannot act on. A structured "no match" tells it to re-inspect and retarget; `internal_error` tells it the server broke, which invites a retry that will fail identically. In a guarded batch this also costs the caller the whole transaction for a condition they could have handled.

Filed rather than fixed because the direct-call half is a public-API shape decision: whether `ReplaceTextRange` should start returning a failed `EditResult` when nothing matched, rather than an empty list, changes an existing contract across the ripple. The batch half — classifying the empty result as something other than `internal_error` — is fixable on its own.

### Suggested resolution

- Give the empty-match case its own `EditErrorCode` (e.g. `TextNotFound`), reported identically on both paths, carrying the anchor and the needle.
[...]

Issue #450, opened 2026-08-14
[Agent] Normalize table, row, column, and cell addressing


JSv4:
Parent: #435

## Goal

Remove the current action-dependent ambiguity in `cellAnchorId` and give every table operation a consistent addressing model.

## Scope notes (revised 2026-08-13 after verifying against `main`)

**The premise is confirmed, and the repository already documents it as a trap.** `tools/mcp-server/ToolCatalog.cs:354` describes one field carrying two anchor kinds: *"insert_row/insert_column/delete_row/delete_column/set_\*: a 'p' (paragraph-inside-the-cell) anchor … replace_cell_content: the cell's own 'tc' anchor instead … these two anchor kinds are not interchangeable."* `docs/architecture/docx_agent_server.md:421-431` carries a section headed **"Anchor-kind trap worth calling out explicitly"** which concludes: *"This is a pre-existing Docxodus API asymmetry, not something this server smooths over."*

**The ambiguity is core-level, not MCP-level.** `DocxSession.cs:5544-5546` gates `ReplaceCellContent` on `Kind != "tc"`; `ResolveCell` (`:7587-7605`), used by the other eleven table ops, instead does `p.Ancestors(W.tc).FirstOrDefault()`. Same parameter name, opposite requirements, on every surface — so the fix cannot be confined to the tool schema.

**This issue now owns table introspection**, previously duplicated in #448's criteria. `Docxodus/DocumentStructure.cs` is the only existing coordinate model and is untested with two incorrect computations — see #471 before building on it.

**Sequence this before #455.** #455 proposes making `MergeCells` emit `w:cellMerge` in tracked mode, which changes which anchors survive a merge — i.e. it rewrites whatever contract this issue establishes. This issue should land first.

## Acceptance criteria

- [ ] Every cell-oriented operation accepts one canonical cell anchor kind.
- [ ] Table, row, column, and cell identities are exposed explicitly in table metadata.
- [ ] Callers can resolve coordinates from anchors and anchors from coordinates.
- [ ] Row/column insertion and deletion return a deterministic mapping of retained, added, and invalidated anchors.
- [x] Merged and vertically merged cells have documented address semantics. — **already satisfied**, see below.
- [ ] Legacy anchor shapes are translated or rejected with a migration-quality error according to compatibility policy.
- [ ] Tool schemas no longer overload one field with action-dependent anchor kinds.
[...]

Issue #435, opened 2026-08-14
[Epic] Agent-safe headless document editing


JSv4:
## Objective

Make headless DOCX editing transactional, guarded, inspectable, and predictable enough for automated document workflows.

## Outcomes

- Atomic mutation batches and genuinely non-mutating previews.
- Optimistic preconditions and retry-safe transaction identities.
- Rich style, layout, table, and inline-content introspection.
- Consistent addressing across document structures.
- First-class operations for hyperlinks, bookmarks, content controls, and images.
- Native revision handling for structural edit families.
- Page-aware citations for rendered document locations.

## Delivery principle

An automated editor should fail closed when its target is stale or ambiguous, should never leave a partially applied transaction unless explicitly requested, and should return enough evidence for a caller to understand exactly what changed.

## Coordination

- **Safety foundation: #447 → #445 → #446, with #449 last.** #447 owns the monotonic document version, which #445, #446, #449 and #454 all consume; #449's own composition criterion names the other three, so it cannot start until they land.
- Read-side introspection (#448) can proceed in parallel and should inform the schemas for the structure-specific edit tracks.
- **#450 before #455.** #455 would make `MergeCells` emit `w:cellMerge`, changing which anchors survive a merge — i.e. rewriting the addressing contract #450 establishes.
- #451, #452 and #453 are independent capability tracks after their addressing/read requirements are clear.
- **#473 blocks #452 and #455.** It recovers a deferral made inside the now-closed #177: a tracked bulk delete spanning a block-level `w:sdt` still hard-removes it. Content-control work cannot claim tracked correctness, and #455's content-control revision family has nothing to enumerate, until it lands.
- #455 can land incrementally by revision family.
- #454 is the only child with a direct rendering dependency on #434.

## Cross-epic dependencies on #436

Several children assume primitives owned by **#436 (Verifiable legal-document delivery)**. These are now referenced in the child issues, and should be scheduled jointly:

| Consumer | Depends on |
|---|---|
| #446 (predicted semantic diff) | #457 — Expand semantic diff beyond block text and kind |
| #453 (semantic diff + package-delta reporting) | #457 and #456 — OOXML package manifests |
| #445 (verifying package topology and relationships were restored) | #456 |

And in the other direction, **#458** (hash-addressed delivery change receipt) consumes #447 and #449 (transaction ids and document versions), #454 (rendered page citations), and #445 (atomic batches, retries, undo/redo in its test matrix).

[...]

PR #491, opened 2026-08-15
Epic #435 acceptance smoke: verify the safety track, fix what it found

### LINKED ISSUE #468: docxodus_mutations preview mode can revert an already-committed edit
Found during an audit of epic #435 against `main` (`f7f4373`). **Code-read, not reproduced — this needs a failing test before it is treated as confirmed.**

## What happens

`docxodus_mutations` with `mode: "preview"` applies every step for real and then calls `Undo()` once per step, as documented at `docs/architecture/docx_agent_server.md:501-506`. The undo count is wrong.

`tools/mcp-server/Dispatcher.cs:655`:

```csharp
mutatingSteps++;
if (succeeded) applied++;
```

`mutatingSteps` is incremented **outside** the `if (succeeded)` branch, so a step that returned `EditResult.Fail` still counts toward the number of undos issued at `:664-668`.

But a failed step records **no** undo entry. Every op's validation returns before `_history.RecordPreOp(...)` — in `ReplaceText`, for example, `DocxSession.cs:4066` (`AnchorNotFound`), `:4069` (`AnchorWrongKind`), `:4073` and `:4086` (payload parse) all precede `:4088`. And `DocxSessionOps.ReplaceText` (`Docxodus/Internal/DocxSessionOps.cs:211-212`) *serializes* that failure to `{"success":false,…}` rather than throwing, so it reaches `:655` instead of the `continue` at `:641`.

Net: a preview batch containing one bad anchor issues **N+1 undos for N recorded entries**. The extra `Undo()` pops a pre-batch snapshot and calls `RestoreSnapshot` (`DocxSession.cs:9224-9227`), rolling the live session back past the batch — silently reverting an edit the caller had already committed.

## Two further preview side effects on live history

1. **Preview destroys the caller's redo stack.** `UndoRing.RecordPreOp` calls `_redo.Clear()` (`Docxodus/Internal/UndoRing.cs:33`). Every preview step clears redo, and it is never restored.
2. **A preview longer than `undoDepth` cannot fully revert.** Default depth is 50 (`DocxSession.cs:1222`); `RecordPreOp` evicts from the front past capacity (`UndoRing.cs:32`); `PopForUndo` returns `(default!, false)` when empty and `Undo()` silently returns `false` (`DocxSession.cs:9225`). A 60-step preview leaves 10 steps permanently applied — and still reports `status: "ok"`.

## Why the existing test does not catch it

`MCP091_Mutations_PreviewMode_LeavesDocumentUnchanged` (`Docxodus.Tests/McpServerDispatcherTests.cs:885-909`) previews a single all-success step and compares only the markdown projection string. It observes no undo/redo depth and exercises no failing step. The `partial` branch at `Dispatcher.cs:670` has no test coverage at all.

## Documentation follow-on

The known-gap notes at `docx_agent_server.md:407-409` and `:501-506` both say Undo is called "once per step that mutated." It is actually called once per step that did not *throw*. That wording should be corrected regardless of how this is fixed.

Related: #446 proposes replacing apply-and-undo preview with an isolated dry run, which would remove this class of bug entirely.

### LINKED ISSUE #471: DocumentStructure table coordinates are untested and computed incorrectly
Found during an audit of epic #435 against `main` (`f7f4373`). Code-read, not reproduced.

`Docxodus/DocumentStructure.cs` is the repository's only table-coordinate model — `RowIndex` / `ColumnIndex` / `RowSpan` / `ColumnSpan` at `:68-83`, and `TableColumnInfo` at `:94-110`. It has **zero tests**, and two of its computations are wrong.

### `RowSpan` is a flag, not a span

`:471-476` assigns `restart → 1`, `continuation → 0`. Its own doc comment at `:78` says "number of rows this cell spans." A cell vertically merged across four rows reports `RowSpan = 1`.

### The column walk ignores `w:gridBefore`

`:417-437` starts each row at `columnIndex = 0` with no handling for `w:gridBefore` (or `w:gridAfter`). In a ragged table — a row that begins partway across the grid — every cell in that row is assigned a column index shifted left by the omitted count, so the reported coordinates do not line up with the table's actual grid.

### Why this matters now

Both #448 (style, formatting and layout introspection) and #450 (table/row/column/cell addressing) specify a table read surface reporting row/column coordinates and spans. `DocumentStructure` is the obvious existing candidate and would be inherited by whichever lands first, so these defects should be fixed or the type explicitly ruled out before either issue starts.

Three further properties make it a poor base as it stands, independent of the two bugs:

- Its identifiers are path-based (`:41`, e.g. `"tbl-0/tr-1/tc-2"`) — an identifier space disjoint from the `kind:scope:unid` anchors every mutation accepts, and shared instead with the annotation subsystem (`AnnotationTarget.cs:13`).
- It operates on document bytes, not a live `DocxSession` (`wasm/DocxodusWasm/DocumentConverter.cs:511`, `npm/src/index.ts:1636`).
- It is absent from the Python and MCP surfaces entirely.

### LINKED ISSUE #472: Five architecture-doc claims describe behavior the code does not implement
Found during an audit of epic #435 against `main` (`f7f4373`).

Five documented behaviors do not exist in the code. Each is load-bearing: a reader triaging the agent-editing epic against these docs would conclude a capability is present when it is not — two of the audit's sub-issues (#452, #453) are specifically at risk of being wrongly marked already-done on the strength of items 1 and 2.

### 1. `docs/architecture/markdown_projection.md:83` — content-control anchor target

> `| w:sdt (content control) | Rendered content, anchor on outer SDT | The SDT itself is an anchor target so callers can address "this content control" |`

`WmlToMarkdownConverter.KindFor` (`:565-577`) emits exactly `p | h | li | tbl | tr | tc | sec | fn | en | cmt` and returns `null` for `w:sdt`. A content control never enters the anchor index and is not addressable.

Related finding in the same area: a body-level `w:sdt` is dropped from the projection entirely. `EmitBlocks` (`:841-865`) dispatches on `W.p` / `W.tbl` / `W.sectPr` with no `else` branch, so it emits neither the control's content nor the `{#unk:…} [unsupported: …]` placeholder used for other unhandled elements — while the paragraphs *inside* it still receive anchors, yielding anchors with no corresponding projection text. `docs/architecture/unsupported_content_placeholders.md` has no `sdt` coverage.

### 2. `docs/architecture/markdown_projection.md:44`, `:82`, `:326-327`, `:443` — image projection

- `:44` lists `img` and `drw` among valid anchor kinds. `KindFor` emits neither.
- `:82` claims `w:drawing` / `w:pict` project as `alt{#img:…}`. No such output path exists — `AppendRunText` (`:1104-1226`) has no `w:drawing` / `w:pict` branch, so an image emits nothing at all into the markdown.
- `:326-327` and `:443` document an `ImageUriBuilder` with a `docxodus://img/{unid}` default. The property is declared at `WmlToMarkdownConverter.cs:111` and is **never read anywhere in the repo** — a dead setting whose XML doc comment asserts a default behavior that does not exist.

### 3. `docs/architecture/docx_mutation_api.md:74` — stale table-CRUD caveat

States that row/column CRUD "v1 assumes a rectangular grid, no `w:gridSpan`". Contradicted by the same document at `:1431-1439`, which gives the span-aware CRUD behavior table, and by the implementation (`DocxSession.cs:7610-7698` grid model).

### 4. `npm/src/session.ts:1038-1039` — invalid `findByKind` tokens

Documents `"row"` and `"cell"` as kind tokens. Neither is valid: the tokens are `tr` and `tc`, and `KindFromToken` throws on anything else (`Docxodus/Ir/IrAnchor.cs:40-74`). This is load-bearing because `tc` lookup is the only documented route to the anchor kind `replace_cell_content` requires (see `tools/mcp-server/ToolCatalog.cs:354`).

### 5. `docs/architecture/docx_agent_server.md:407-409` and `:501-506` — preview undo count

Both say `docxodus_mutations` preview calls Undo "once per step that mutated." It is called once per step that did not *throw*, including steps that returned `success: false` without recording an undo entry. Tracked separately as #468; the wording needs correcting in both places regardless of how that is resolved.

JSv4:
Runs the epic #435 acceptance gate end to end against a real document, verifies and closes the three issues that gate depended on, and fixes the one engine defect the run turned up.

Closes #468. Closes #471. Closes #472.

## The smoke

A committed, re-runnable MCP workflow against `TestFiles/NVCA-Model-COI.docx` — 44 OPC parts, 8 headers, 10 footers, 94 footnotes, 392 bookmarks, 462 anchors, 4 sections.

| Pass | Calls | Assertions | Failed | Expected failures | Replay mismatches |
|---|---:|---:|---:|---:|---:|
| Workflow | 45 | 81 | **0** | 4 | **0** |
| Reopen validation | 13 | 27 | **0** | — | — |

It walks inspect → isolated preview → atomic apply under a transaction id → retry → audit, then an intentional stale request and an intentional atomic failure, a page citation, and save/close/reopen. Anchors are discovered by content, never assumed. The applied `resultVersion`, semantic deltas and per-step affected anchors are asserted *against the preview's captured predictions* rather than literals, and the retry is compared **byte-exact at the wire level**.

`packageHash` differs between preview and apply, which is the documented contract rather than a defect — the batch generates comment/footnote ids and revision timestamps, and both receipts carry the warning saying equivalence is semantic. Equivalence is verified where it is promised and deliberately not asserted where the receipt says not to. The same discipline applies to created anchors: counts are compared, ids are not.

Four calls are *supposed* to fail, and the harness treats a success there as the defect — a bookmark and a table insert refused under tracked recording, a stale precondition, and the batch that rolls back. Failing closed is the property under test.

Full write-up, evidence and reproduction: [`docs/architecture/epic_435_acceptance_smoke.md`](docs/architecture/epic_435_acceptance_smoke.md).

## The defect it found — fixed

Under `render_inline`, an agent could not re-find the edit it had just made. `replace_text_range` succeeded and the projection showed the new text, but `docxodus_search` returned nothing, `apply_format_by_substring` reported `offset_out_of_range`, and a follow-up surgical op could not address it.

`InlineRuns` descends only into the containers in `InlineContainerNames`, and `w:ins` was not one of them, so every run inside a tracked insertion was missing from the flat string all offset-addressed ops share — while `Project().Markdown` and `TextPreview` included it.

Not limited to text a session authored: an incoming Word redline had its inserted spans skipped too. `RA001-Tracked-Revisions-01.docx` holds six occurrences of "chromatogram" and search returned exactly the two outside `w:ins`.

The split is visible text versus text the document says was removed, not plain versus revised runs: `w:ins`/`w:moveTo` hold ordinary `w:t` that a reader sees, `w:del`/`w:moveFrom` hold `w:delText` that `RunText` never reads. The first pair is now transparent, the second deliberately is not. `DS409` covers the round trip through Grep, TextPreview and a follow-up surgical op; `DS410` covers pre-existing insertions and a move destination, and pins that deleted text stays unfindable.

## Issue verdicts

**#468** — the bug class is gone structurally, not merely untested: `mode: preview` routes through an isolated shadow clone and the dispatcher's only remaining `Undo()` is the explicit edit action. `MCP098` already covered the `partial` branch. The gap was that it previews from version 0, where a surplus undo is a no-op, so **`MCP150`** reproduces the reported scenario at the MCP surface — a committed edit, a seeded redo cursor, `undoDepth: 1`, and two previews ending in a bad anchor. Re-injecting the defect turns `MCP150` and `MCP148` red while `MCP091`/`MCP098` stay green, which is how the new test is shown not to be vacuous.

**#471** — both coordinate defects are fixed. #450 rebased `DocumentStructure`'s table walk onto `TableGridModel`, so `ColumnIndex` honours `w:gridBefore` and `RowSpan` is a span rather than a restart/continuation flag, with `DT251` testing a ragged spanned grid. The doc comments still described the old flag behaviour and are corrected.

**#472** — split verdict, all five now resolved. Items 1 and 5 were already fixed by the merged work; items 2, 3 and 4 were still wrong. Images are genuinely not projected (verified: no `img`/`drw` kind is ever assigned and no image markup is emitted), so the docs now describe that instead of promising it; the rectangular-grid CRUD caveat had moved to `editor_ui_surface.md`; npm `findByKind` still documented `"row"`/`"cell"` instead of `tr`/`tc`. One stale preview-undo sibling was also found in `docx_mutation_api.md`.

## Filed, not fixed

**#490** — `replace_text_range` matching nothing returns `[]` from a direct call (indistinguishable from success) but `internal_error` inside a batch. Making the direct call report it is a public-API shape change, so it is filed with a suggested resolution rather than changed here. It does not block the gate.

## Harness changes

`mcp_probe.py` gains `expectFailure` (which also flags an *unexpected success*, so a negative probe cannot pass vacuously once the guard it exercises stops guarding) and `expectSameAs` (raw wire-text comparison, since byte-exact replay is what a transaction promises). `expect` values now substitute `$variables`. The trace no longer dies with the run — an unresolvable assertion path is a failed assertion and an unresolvable capture is a recorded stop, because the trace is the evidence.

The workflow JSON is generated rather than hand-maintained: the preview, apply and retry must send a byte-identical step array or the retry is a `transaction_conflict` instead of a replay.

## Verification

- .NET suite: **3793 passed, 0 failed, 3 skipped**
- Both smoke passes green, as tabled above
- `npx tsc --noEmit` clean
- Package integrity on the saved output: zip valid, 44 → 45 parts (only `word/comments.xml` added), no part lost, no dangling relationships, one external relationship

Not covered: no third-party render oracle — `soffice --convert-to pdf` did not finish within 15 minutes on the 610 KB tracked-changes output, so integrity was established structurally and by a full reopen instead. The Playwright suite is running at time of writing and will be reported in a follow-up comment; the only npm change in this PR is a JSDoc correction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ Docxodus.Tests/Ir/Diff/FuzzArtifacts/
 
 # macOS
 .DS_Store
+tools/mcp-server/smoke/__pycache__/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -405,13 +405,16 @@ All notable changes to this project will be documented in this file.
   so a search over an incoming redline silently missed every inserted span.
   `w:ins`/`w:moveTo` now join the transparent inline containers; `w:del`/`w:moveFrom`
   deliberately do not, because their content is `w:delText` and deleted text is not
-  visible text. A note/endnote insertion whose requested offset is strictly inside one of
-  those revision wrappers now fails closed with `unsupported_inline_boundary`; the citation
-  inserter cannot split revision semantics and previously reported success after silently
-  placing the reference at the wrapper's far edge. Revision-edge offsets remain exact.
-  Coverage: `DocxSessionSurgicalTrackedChangesTests` DS409-DS410 and
-  `DocxSessionNoteAuthoringTests` DS339-DS340. Found by the issue #435 acceptance smoke and
-  its PR #491 adversarial review (`tools/mcp-server/smoke/epic-435-workflow.json`).
+  visible text. Making that text addressable also made it a valid target for ops that insert
+  or partition a paragraph's top-level children, and those cannot split a revision wrapper —
+  so a note/endnote citation and `SplitParagraph` whose requested offset is strictly inside
+  `w:ins`/`w:moveTo` now fail closed with `unsupported_inline_boundary`. Both previously
+  reported success after silently landing at the wrapper's far edge. Offsets exactly at a
+  wrapper's edge remain exact, and image insertion already refused this case.
+  Coverage: `DocxSessionSurgicalTrackedChangesTests` DS409-DS410,
+  `DocxSessionNoteAuthoringTests` DS339-DS340, and `DocxSessionPR131RegressionTests`
+  DS095-DS096. Found by the issue #435 acceptance smoke and its PR #491 adversarial review
+  (`tools/mcp-server/smoke/epic-435-workflow.json`).
 - **An ordinary Word picture is no longer reported read-only because of a rendering
   hint.** The blip-extension guard that refuses to replace a picture holding a second
   image payload (an SVG `asvg:svgBlip` beside its raster fallback, or an artistic-effects
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -394,6 +394,19 @@ All notable changes to this project will be documented in this file.
   version bump at release time.
 
 ### Fixed
+- **Tracked insertions are part of a paragraph's visible text, so an agent can re-find
+  the edit it just made.** Under `TrackedChangeMode.RenderInline`, text a mutation
+  inserted landed inside `w:ins`, and `InlineRuns` — the shared walk behind the flat
+  string that every offset-addressed op works over — did not descend into it. The result
+  was a split brain on a document the caller had just edited: `Project().Markdown` and the
+  anchor's `TextPreview` showed the new text while `Grep`/`docxodus_search` returned
+  nothing, `ReplaceTextRange` could not address it, and `apply_format_by_substring`
+  reported `offset_out_of_range`. Insertions arriving from Word were skipped the same way,
+  so a search over an incoming redline silently missed every inserted span.
+  `w:ins`/`w:moveTo` now join the transparent inline containers; `w:del`/`w:moveFrom`
+  deliberately do not, because their content is `w:delText` and deleted text is not
+  visible text. Coverage: `DocxSessionSurgicalTrackedChangesTests` DS409-DS410. Found by
+  the issue #435 acceptance smoke (`tools/mcp-server/smoke/epic-435-workflow.json`).
 - **An ordinary Word picture is no longer reported read-only because of a rendering
   hint.** The blip-extension guard that refuses to replace a picture holding a second
   image payload (an SVG `asvg:svgBlip` beside its raster fallback, or an artistic-effects
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -405,8 +405,13 @@ All notable changes to this project will be documented in this file.
   so a search over an incoming redline silently missed every inserted span.
   `w:ins`/`w:moveTo` now join the transparent inline containers; `w:del`/`w:moveFrom`
   deliberately do not, because their content is `w:delText` and deleted text is not
-  visible text. Coverage: `DocxSessionSurgicalTrackedChangesTests` DS409-DS410. Found by
-  the issue #435 acceptance smoke (`tools/mcp-server/smoke/epic-435-workflow.json`).
+  visible text. A note/endnote insertion whose requested offset is strictly inside one of
+  those revision wrappers now fails closed with `unsupported_inline_boundary`; the citation
+  inserter cannot split revision semantics and previously reported success after silently
+  placing the reference at the wrapper's far edge. Revision-edge offsets remain exact.
+  Coverage: `DocxSessionSurgicalTrackedChangesTests` DS409-DS410 and
+  `DocxSessionNoteAuthoringTests` DS339-DS340. Found by the issue #435 acceptance smoke and
+  its PR #491 adversarial review (`tools/mcp-server/smoke/epic-435-workflow.json`).
 - **An ordinary Word picture is no longer reported read-only because of a rendering
   hint.** The blip-extension guard that refuses to replace a picture holding a second
   image payload (an SVG `asvg:svgBlip` beside its raster fallback, or an artistic-effects
--- a/Docxodus.Tests/DocxSessionNoteAuthoringTests.cs
+++ b/Docxodus.Tests/DocxSessionNoteAuthoringTests.cs
@@ -554,6 +554,106 @@ public class DocxSessionNoteAuthoringTests
     }
 
     /// <summary>
+    /// Regression found while reviewing PR #491. Once tracked insertions joined the visible-text
+    /// walk, note offsets could resolve inside <c>w:ins</c>/<c>w:moveTo</c>. The run splitter cut
+    /// the nested run, but the top-level inserter treated the revision wrapper as atomic and
+    /// silently placed the citation after the entire revision. Refuse that boundary before the
+    /// history snapshot instead of reporting a successful wrong-location edit.
+    /// </summary>
+    [Theory]
+    [InlineData("ins", true)]
+    [InlineData("moveTo", false)]
+    public void DS339_InsertNote_InsideVisibleRevision_FailsClosedWithoutMutation(
+        string revisionName, bool footnote)
+    {
+        using var session = new DocxSession(BuildDocWithVisibleRevision(revisionName));
+        var anchor = FirstBodyParagraph(session);
+        var before = session.GetPackageContentHash();
+        var version = session.Version;
+        var undoCount = session.UndoCount;
+        var redoCount = session.RedoCount;
+
+        var result = footnote
+            ? session.InsertFootnote(anchor, 4, "Must not be misplaced.")
+            : session.InsertEndnote(anchor, 4, "Must not be misplaced.");
+
+        Assert.False(result.Success);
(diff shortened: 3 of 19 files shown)


JSv4:
## Browser suite result

**518 passed, 5 failed, 9 skipped.** The 5 failures are pre-existing and environmental, not caused by this PR.

All five are in `tabs-visual.spec.ts` — pixel-comparison tests against committed PNG baselines — and every one fails on a screenshot *height* mismatch (e.g. `HC024`: expected 896×443, received 896×472), which is the signature of local font metrics differing from the machine that recorded the baselines.

Two independent checks that this PR is not the cause:

1. **The fixtures cannot reach the change.** `HC022`, `HC024`, `HC025`, `HC026` and `HC027` contain **zero** `w:ins` and zero `w:moveTo` elements. `InlineRuns` only iterates differently when one of those is present, so on these documents it enumerates byte-identically to before.
2. **They fail the same way without the change.** Reverting `Docxodus/DocxSession.cs` to `origin/main`, rebuilding the WASM bundle and re-running that spec reproduces exactly the same 5 failures (`5 failed, 1 passed`).

Everything that exercises the changed code path is green, including the tracked-changes, annotation, batch and session specs.

### A note on how this was verified

My first two attempts at this run were invalid and I want that on the record rather than buried:

- The first used `npx playwright test`, which bypasses npm's `pretest` hook — the step that copies the harness HTML pages and bundles into `dist/wasm/`. Every page 404'd and the run died at 50 of 532 tests. It measured nothing.
- The second hit a stale `cors-server.py` holding port 8083 from the first run's leftovers, so the web server never started.

The 518/5/9 figures above come from the third, correct run via `npm test`.

### Full verification for this PR

| Check | Result |
|---|---|
| .NET suite | 3793 passed, 0 failed, 3 skipped |
| Epic smoke workflow | 45 calls, 81 assertions, 0 failures, `replayMismatches: 0` |
| Reopen validation | 13 calls, 27 assertions, 0 failures |
| Playwright | 518 passed, 5 pre-existing environmental failures, 9 skipped |
| `tsc --noEmit` | clean |
| Saved-package integrity | zip valid, 44 → 45 parts, no part lost, no dangling relationships |

Still not covered: no third-party render oracle — `soffice --convert-to pdf` did not finish within 15 minutes on the 610 KB tracked-changes output.

PR #494, opened 2026-08-15
Add stable full-surface semantic DOCX changes

JSv4:
Closes #457

Stacked on #493 / issue #456. Review the semantic-diff commits in this PR; once #493 lands, this branch can be retargeted to `main` without changing the #457 patch.

## Summary

- adds a stable, versioned `docxodus.semantic-changes` v1 schema with deterministic ordering and canonical JSON
- reports text/structure, formatting/styles, lists/numbering, tables, sections/page setup, stories/notes/comments, fields/links/bookmarks/controls, media/relationships, revisions/annotations, and opaque package changes
- distinguishes insert/delete/move/modify and supplies owning part, paths, side-specific anchors/scopes, and typed before/after values
- exposes the same semantic evidence through .NET, session, WASM/npm, Python, and MCP surfaces without changing legacy diff/redline formats
- publishes and tests the formal JSON Schema plus documented performance/output budgets

## Architecture and safety

- keeps the existing IR/edit-script implementation as the modeled alignment authority
- supplements only the surfaces IR does not model using #456's manifest relationships, digests, locations, XML normalizer, and content-type resolution
- performs mandatory manifest preflight before SDK/IR parsing, even when package supplementation is disabled
- enforces declared and actual per-entry, aggregate, XML, and compression-ratio limits
- shares one bounded package pass and retains detached parsed XML only—no second ZIP reader, archive handles, or duplicate raw payload copies
- suppresses relationship-ID churn and XML serialization-only changes while preserving opaque XML whitespace/comments/processing instructions

## Verification

- `SemanticDiffTests`: 30 passed
- `PackageManifestTests`: 30 passed
- `McpServerDispatcherTests`: 85 passed
- Python semantic host tests: 6 passed
- npm TypeScript typecheck: passed
- .NET build: 0 errors
- `git diff --check`: passed

Measured guards:

- 1,000 paragraphs: 132.4 ms, 1 change, 589 canonical bytes
- 200 tables / 400 rows: 205.0 ms, 8 changes, 6,663 canonical bytes

Issue #464, opened 2026-08-14
[Verify] Prove generated redline accept/reject reversibility


JSv4:
Parent: #436

## Goal

Turn the redline round-trip invariant into an explicit delivery proof.

For a generated redline, accepting its generated changes should yield the intended final document, while rejecting them should yield the selected baseline.

## Acceptance criteria

- [ ] Verification identifies which revisions belong to the generated comparison/edit set versus pre-existing revisions.
- [ ] Accepting the generated revision set is compared with the intended final document.
- [ ] Rejecting the generated revision set is compared with the selected baseline document.
- [ ] Proof includes block text, formatting, lists, tables, sections, headers/footers, notes, comments, bookmarks, and supported structural content.
- [ ] Modeled semantic equivalence is distinguished from byte-for-byte package equivalence.
- [ ] Unknown/unmodeled package differences are reported rather than silently ignored.
- [ ] Failures identify the first divergent part/anchor and the applicable revision IDs.
- [ ] Tests cover pre-existing revisions, moves, formatting revisions, tables, notes, comments, and multi-author comparisons.
- [ ] The proof result is embeddable in the delivery change receipt.

Issue #463, opened 2026-08-14
[Verify] Add a comprehensive deliverable validator


JSv4:
Parent: #436

## Goal

Provide one structured `verifyDeliverable` operation that determines whether a DOCX and its companion artifacts are safe to return.

## Validation areas

- ZIP/package integrity and content types.
- Open XML validation error delta.
- Relationship and media closure.
- Bookmark, comment, note, revision, content-control, list, and field consistency.
- Remaining template placeholders.
- Unsupported renderer content and font substitutions.
- Expected versus actual package/semantic delta.
- Companion PDF/HTML page and artifact metadata.

## Acceptance criteria

- [ ] Validation returns structured findings with stable codes, severity, owning part, anchor where available, and remediation guidance.
- [ ] Pre-existing input problems are distinguished from newly introduced problems.
- [ ] Broken or dangling relationships and missing media are detected.
- [ ] Comment/revision/note markers and definitions are checked for paired/reachable consistency.
- [ ] Remaining placeholders and unresolved workflow markers are reported.
- [ ] An optional policy fails on unexpected semantic or package changes.
- [ ] Render warnings and missing/substituted fonts are included when companion artifacts are supplied.
- [ ] Strict, standard, and report-only policy presets are documented.
- [ ] Tests include intentionally corrupted and pre-invalid documents as well as valid complex fixtures.

Issue #457, opened 2026-08-14
[Verify] Expand semantic diff beyond block text and kind


JSv4:
Parent: #436

## Goal

Report meaningful document changes that the current anchor-keyed text/kind diff cannot see.

## Change families

- Text and block structure.
- Paragraph/run formatting and styles.
- Lists and numbering.
- Tables, rows, cells, spans, widths, and styling.
- Sections, page setup, headers/footers, fields, notes, and comments.
- Hyperlinks, bookmarks, content controls, images/media, and relationships.
- Revisions and annotations.

## Acceptance criteria

- [ ] Diff output has stable, versioned JSON schemas and deterministic ordering.
- [ ] Every change identifies its owning part, affected anchors, change family, and before/after values.
- [ ] Insert/delete/move/modify operations are distinguishable where identity permits.
- [ ] Formatting-only and relationship-only changes are visible.
- [ ] Serialization-only package differences are not misreported as semantic changes.
- [ ] Unknown/unmodeled part changes remain visible as opaque package changes.
- [ ] Existing text/kind diff formats remain backward compatible or receive a documented version transition.
- [ ] Tests cover each change family and mixed multi-part edits.
- [ ] Large-document performance and output-size bounds are measured and documented.

PR #497, opened 2026-08-16
Add hash-addressed delivery change receipts

### LINKED ISSUE #458: [Verify] Generate a hash-addressed delivery change receipt
Parent: #436

## Goal

Emit a portable machine-readable record connecting the source document, requested operations, resulting document, and verification evidence.

## Receipt contents

- Schema/version identifiers.
- Input/output document hashes.
- Transaction IDs and document versions.
- Normalized requested operations.
- Resolved anchors, scopes, and rendered page citations where available.
- Semantic before/after changes.
- OOXML parts and relationships changed.
- Revision/comment changes and authorship.
- Warnings, validation results, and renderer fingerprint.
- Artifact hashes for generated DOCX, HTML, PDF, and reports.

## Acceptance criteria

- [ ] Receipt serialization and hashing are deterministic.
- [ ] Every committed transaction can contribute a receipt entry.
- [ ] The receipt distinguishes user-requested changes, derived changes, and unexpected changes.
- [ ] It contains enough identifiers to verify all listed output artifact hashes independently.
- [ ] Sensitive full document text is optional; hash/summary-only modes are supported.
- [ ] Page citations are explicitly tied to the document version and render fingerprint that produced them.
- [ ] Schema evolution and backward-compatibility policy are documented.
- [ ] Tampering with a receipt or referenced artifact is detected.
- [ ] Tests cover single edits, atomic batches, retries, undo/redo, redline generation, and multi-artifact delivery.

JSv4:
## Summary

- add deterministic, hash-addressed delivery change receipts over source and delivered package identities
- retain normalized transaction, attribution, authored-change, lineage, privacy-profile, PageMap, artifact, and tamper-verification evidence
- consume #457 canonical semantic evidence, #463 shared package deltas and deliverable reports, and #464 canonical reversibility proofs
- enforce bounded parsing/serialization and portable artifact paths with fail-closed schema, lineage, digest, and resource validation
- distinguish exact idempotent retries from conflicting results while preserving distinct identified and anonymous no-op transactions

## Stack

**Rebased onto `main`; targets `main` directly.** Every dependency has landed — #493, #494, and
#496 are all in `main` — so this branch no longer carries their commits. It previously replayed
18 commits that existed elsewhere (8 of #456/#457, 5 of #463, 6 of #464), which is why its diff
read as +40,189 against a merge-base predating the whole stack.

`git rebase --onto origin/main <last-inherited-commit>` dropped all 18 with **zero conflicts**,
leaving the three commits that are actually this PR's work:

| | Before | After |
|---|---|---|
| Commits | 21 | **3** |
| Diff | +40,189 / −372 | **+11,891 / −16** (21 files) |
| Mergeable | conflicting | **clean** |

The inherited #464 commits it had been carrying were the *older* lineage — the one predating the
#515 `delText` fix and the fixture sweep. Taking main's versions rather than replaying them keeps
that work intact.

## Verification

Run against the rebased tree:

- full suite: 4,236 passed / 0 failed / 3 pre-existing skips
- `dotnet build`: 0 errors
- rebase produced no conflicts in any of the three replayed commits

The rebase was verified against the #496 branch head, then re-parented onto main's squash of it
(`4ee78d5`) once that merged. The two trees are byte-identical, so the verification above applies
to the pushed commits unchanged.

Earlier stack-local results, superseded by the run above: 179 passed in the Release
dependency-stack run, 34 passed in the focused delivery-receipt run, independent post-restack
review approved with no blocker/high/medium findings, covering shared target-aware package
ordering, exact retry evidence, distinct no-op identities, duplicate revision IDs across parts,
and real #463/#464 canonical artifacts.

Closes #458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,25 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- **Deterministic delivery change receipts (#458).** Added a versioned, canonical JSON
+  receipt that binds source/delivered package identities to every supplied mutation transaction,
+  normalized requests, explicit undo/redo lineage, requested/derived/unexpected package-change
+  attribution, required typed semantic evidence, optional validation/reversibility evidence, and
+  independently hashed clean DOCX, review DOCX, HTML, PDF, image, and report artifacts. Exactly
+  one clean DOCX must match the delivered raw package bytes, while exact #457 canonical semantic
+  bytes cover the source-to-delivered comparison and every state-changing transaction. Privacy
+  profiles support hash-only, structural-summary, and full-evidence output. Page citations are
+  accepted only when their exact PageMap bytes project the claimed coordinates for a reachable
+  document version/package digest and match the render fingerprint and artifact. Undo/redo is
+  checked as a LIFO state machine, and any indexed package attribution may point only to a
+  successful retained operation. Failed atomic receipts retain the complete requested-operation
+  list with explicit not-executed/rolled-back status, while a genuinely successful no-op may retain
+  an empty edit-result list. Strict typed semantic reconstruction, authoritative clean-DOCX inventory
+  recomputation, portable path normalization, canonical collection-order checks, aggregate object
+  budgets, stream-charged JSON collections, and capped canonical writers fail closed under forged or
+  oversized input. The portable verifier detects receipt,
+  record, artifact, semantic, lineage, and citation-binding tampering. See
+  [`docs/architecture/delivery_change_receipt.md`](docs/architecture/delivery_change_receipt.md).
 - **Deterministic, non-mutating DOCX package manifests (#456).** Added a versioned
   verification artifact that inventories every OPC entry and content type, preserves duplicate
   occurrences, resolves every package/part relationship, reports dangling references and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -412,6 +412,14 @@ All notable changes to this project will be documented in this file.
   recompile.
 
 ### Changed
+- **A batch step whose mutation records no edits is now a successful no-op (#458).**
+  `MutationBatchStep.Mutation` may return an empty edit-result list; previously the
+  step failed with `internal_error` and rolled an atomic batch back. This means an
+  edit that matches nothing (for example a `replace_text_range` whose search text
+  is stale) no longer blocks the rest of the batch — callers that need "this edit
+  must land" semantics should pass `ExpectedMatchCount` (or a preflight) so a
+  zero-match step fails explicitly instead of vacuously succeeding. Delivery
+  receipts record such steps as portable no-op evidence.
 - **A footnote too long for its page now keeps flowing instead of being clipped.**
   Browser pagination reserves at most 60% of the body height for the note band; a
   note whose tail did not fit there used to be handed to the next page whole and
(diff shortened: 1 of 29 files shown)


JSv4:
## Review fix round

An adversarial multi-agent review of this PR surfaced 15 confirmed findings plus a duplicated-contract cluster. All are addressed in the six commits ending at 2fbc86f, each landed test-first (12 new regression tests; full suite 4,248/0, npm Playwright 522 passed with only the pre-existing environment-sensitive `tabs-visual` snapshots differing locally).

**Determinism / tamper evidence**
- Annotation `created` stamps parsed to `Kind=Local` and were serialized with the machine's UTC offset inside the digest-covered payload, making `ReceiptDigest` timezone-dependent. The custom-XML parse now normalizes to UTC and every wire timestamp goes through one `UtcRoundtrip` helper (DCR046).
- A forged artifact registered under `Role=SemanticDiff` (or any typed-evidence role) that nothing referenced bypassed byte-exactness and verified `IsValid=true`. Unbound evidence-role artifacts now fail closed at build **and** verify (`unbound_evidence_artifact`, DCR047).
- `IsExactCanonical` accepted undefined integer enum values (`"family":9999`) because `allowIntegerValues` defaulted true; all wire enum converters now reject integers (RP051, DCR042).
- The verifier computed and discarded the normalized artifact path, accepting backslash paths the builder can never emit (DCR043), and emitted non-snake-case codes (`artifact_digestmismatch`); both fixed (DCR008).

**Builder-accepts-verifier-rejects traps**
- Idempotent previews (the engine bumps the version with byte-identical content) collapse to the canonical no-op prediction at intake (DCR049); entry versions/ids derive from the contribution identities.
- Every failed batch-step result must carry a structured edit error, not just the first failure (DCR050).
- Lineage-event document identities are validated at `AddLineageEvent` via a shared `ValidateDocument` (DCR051).
- An exact semantic-change-set re-registration is an idempotent retry instead of a duplicate that `Build()` later misreports as *missing* evidence (DCR052).
- Renders bound to the in-receipt review copy's identity now verify; fabricated identities fail at build (`IsArtifactDocumentReachable`, DCR048).

**Robustness / parity**
- Secondary `JsonDocument.Parse` calls used the default depth 64 under a 128 pipeline limit — legally-deep receipts crashed with raw `JsonException`; all bounded by `DeliveryReceiptLimits.MaxAllowedJsonDepth` (DCR044).
- One stale citation poisoned the page-map verdict cache for later, correctly-bound citations sharing the artifact (DCR045).
- The npm batch runner still failed empty mutation-result lists that core treats as successful no-ops; aligned, with a Playwright parity spec. The zero-match-step behavior change is now called out in the CHANGELOG (`ExpectedMatchCount` for must-land semantics).
- The core-only surface scope is documented in the architecture doc; the cross-surface ripple is tracked in #520.

**Duplicated-contract cluster**
- New `DeliveryReceiptContract` single-owner for the evidence-kind→schema map (which had already drifted: throw vs `string.Empty`), scope parser, sha256 token grammar, EditResult anchor arity, redacted free-text format/matcher, and citation coordinate equality. Digest comparison collapses to shared `DigestEquals`/`OptionalDigestEquals` (PackageDelta's case-insensitive variant was dead tolerance). Also: clean-DOCX manifest reused between registration and `Build()`, artifact digests memoized across verify passes, and receipt creation no longer serializes the whole envelope a second time just to measure it.

Issue #489, opened 2026-08-15
Footnote continuation stops at the paragraph: an oversized note paragraph is clipped, not split


JSv4:
Follow-up from #488, which pinned the note-drain behaviour into `pagination-keep-with-next`.

The rule we want is the one you'd expect: a footnote that does not fit its page's note band puts what fits on that page and pushes the rest to the next, and if the rest does not fit *there* it pushes again — adding pages past the end of the body text if that is what it takes to render the whole note.

**Most of that already works.** I verified the cascade against `npm/dist/pagination.bundle.js` at `6296c7d`, on a 100pt body with a 60pt note band (`MAX_FOOTNOTE_AREA_RATIO = 0.6`) and 30pt note paragraphs:

| Case | Result |
|---|---|
| One note, 6 paragraphs, cited on the first block | 6 pages, one paragraph per page — cascades correctly |
| Same, cited on the **last** body block | 6 pages — keeps adding note-only pages past the body |
| **Two** notes on one page, each needing a split | 6 pages, `a1 a2 a3` then `b1 b2 b3` — nothing dropped |

So the recursion, the note-only pages after the body ends, and the two-notes-both-split case (the single continuation slot backed by the deferral queue) are all sound. This issue is only about the case where the cascade cannot engage.

## The gap: the split is only ever taken at a paragraph boundary

`splitContinuationForPage` and `splitFootnoteToFit` partition at whole-child boundaries and deliberately always advance by at least one element, so a paragraph taller than the band is emitted alone and the band clips it. That is the documented "established clipped fallback", and for a note that is one long paragraph it means most of the note is silently gone.

**Case C — one 150pt paragraph, 60pt band:**

```
page 1: body=[lead1,body]  notes="f1huge"  noteHeight=80px  overflow:hidden
total pages: 1
```

~200px of paragraph rendered into an 80px `overflow: hidden` box. The remainder is not on a later page — it is not anywhere. No second page is created, because from the engine's point of view the continuation was fully consumed.

**Case C2 — a 150pt paragraph followed by two 30pt ones — is worse:**

```
page 1: body=[lead1,body]  notes=null            <-- no note area at all, though the citation is here
page 2: body=[]            notes="huge after1 after2"  noteHeight=80px  overflow:hidden
total pages: 2
```

[...]

Issue #444, opened 2026-08-14
[PDF] Add explicit revision and comment rendering profiles


JSv4:
Parent: #434

## Goal

Let callers produce unambiguous final, original, and review PDFs from documents containing native revisions and comments.

## Proposed profiles

- `final`: show the accepted-result view.
- `original`: show the rejected-result view.
- `markup`: show supported inserted, deleted, moved, and formatting changes.
- Comment visibility remains an orthogonal option such as hidden, inline, endnote-style, or margin where supported.

## Acceptance criteria

- [ ] Each profile has documented, deterministic semantics.
- [ ] Insertions, deletions, moves, formatting changes, authors, and dates render according to profile.
- [ ] Comment bodies, ranges, replies, authors, and resolved state follow the selected comment policy.
- [ ] Existing revisions are never silently accepted, rejected, or removed from the source DOCX.
- [ ] PDF text extraction behaves predictably for visible inserted and deleted content.
- [ ] Tests cover overlapping revisions/comments and revisions in body, headers/footers, footnotes, and endnotes where supported.
- [ ] Unsupported revision families produce explicit warnings.
- [ ] The same profile vocabulary is used by standalone HTML, PDF API, and CLI.

Issue #442, opened 2026-08-14
[PDF] Define and implement the font runtime contract


JSv4:
Parent: #434

## Goal

Make font resolution observable and reproducible so layout differences are diagnosable rather than silent.

## Acceptance criteria

- [ ] The CLI supports one or more caller-provided font directories.
- [ ] Export records requested families, resolved families, substitutions, and missing fonts.
- [ ] Missing fonts produce structured warnings or errors according to an explicit strictness option.
- [ ] The renderer fingerprint includes the layout-relevant Chromium and font-configuration identities.
- [ ] The supported CLI distribution has a documented reproducible font configuration.
- [ ] Library callers can inject their own browser/font environment without importing a bundled browser into the browser/WASM package.
- [ ] Tests cover exact match, intentional substitution, missing family, font load failure, and metrically different substitution.
- [ ] The existing visual-parity font contract is reused or generalized rather than duplicated.

PR #501, opened 2026-08-16
Materialize standalone paginated HTML

### LINKED ISSUE #438: [HTML] Materialize standalone paginated HTML
Parent: #434

## Goal

Expose a supported API that converts DOCX bytes into a finalized, offline paginated HTML artifact.

Today paginated conversion produces staging markup and metadata consumed by the TypeScript pagination engine. This issue turns that pipeline into a portable artifact rather than requiring every caller to reconstruct the orchestration.

## Proposed surface

`convertDocxToPaginatedHtml(document, options)` should return the finalized HTML plus useful render metadata such as page count, page map, warnings, and renderer fingerprint.

## Acceptance criteria

- [ ] The returned artifact contains the finalized page-box tree, not an unprocessed staging tree.
- [ ] The artifact opens from disk without network access.
- [ ] Document CSS, supported images, charts, SVG, headers, footers, footnotes, endnotes, revisions, comments, and anchor metadata survive materialization.
- [ ] Viewer-only scale, gaps, shadows, and host margins do not affect print geometry.
- [ ] Text remains selectable and searchable.
- [ ] Reopening the artifact produces the same logical page count and section geometry as the original materialization.
- [ ] The result includes structured warnings for missing fonts and unsupported content.
- [ ] Public types, npm documentation, examples, and Playwright coverage are added.

JSv4:
Closes #438.

Depends on #500 (merged); rebased onto `main`.

## Summary

- adds the UI-free `docxodus/export-browser` materializer for complete, standalone paginated HTML
- snapshots caller bytes, validates the #493 package manifest/digest, enforces bounded resource and phase limits, and emits typed success/failure reports
- isolates conversion and two independent pagination passes, verifies determinism, sanitizes automatic resources and active content, normalizes IDs/fragment links, and reopens the serialized document offline before returning it
- binds the HTML, final-tree PageMap, renderer/runtime fingerprint, resource inventory, font observations, and report with SHA-256 digests
- fixes owner-part relationship resolution for header/footer/note/comment images, charts, and hyperlinks
- makes pagination use the supplied element's DOM realm and adds cooperative deadline/cancellation checks
- packages a closed 46-asset runtime graph, schemas, documentation, and a browser example

## Verification

- `npm run build`
- `npm run pretest` — 163 package-boundary files
- Playwright integration gate — 27 passed
- `PaginatedHeaderFooterContentTests` — 3 passed
- HTML converter corpus — 53 passed
- independent final CSS tokenizer/sanitizer audit — clean

The Playwright workflow now uploads `npm/playwright-report/` and `npm/test-results/` with `if: always()`. Successful runs include directly viewable standalone HTML, PageMap/report JSON, request logs, offline-reopen screenshots, and a self-contained artifact viewer; expected strict/clipping failures retain their structured failed reports too.

## Intentional follow-ups

- original-profile fidelity remains fail-closed pending #444
- strict font verification remains fail-closed pending #442
- oversized indivisible note/body flows remain fail-closed pending #489


JSv4:
## Review

Rebased onto `main` and re-scoped: this PR was showing 8,000+ lines because it was branched off `agent/issue-437-export-architecture`, carrying a stale pre-review copy of the already-merged #493 manifest work. The genuine payload is **23 files, +5,702/−133**, and that size is justified — it maps directly onto the contract #500 just froze (closed schema v1, manifest validated as a closed schema, CSS/URL sanitization, offline-reopen verification, canonical JCS, fingerprint, typed error taxonomy, phased readiness) plus #438's eight acceptance criteria. None of the new machinery duplicates anything already in `npm/src`.

Quality is high where it matters: `sanitizeCss` retokenizes after every rewrite and **fails closed** (drops all CSS) rather than emitting unproven output; `assertStandaloneResourceAudit` independently re-proves the serialized document instead of trusting the sanitizer; there is not a single swallowing `catch` in the new code.

Findings below are follow-ups, not merge blockers.

### 1. Typed error taxonomy depends on unpinned worker message wording (durability)

`export-browser.ts:3132` and `:3176` classify worker failures by string-matching their messages:

```ts
if (String(error).includes("exceeds compressedDocxBytes")) {
```

The de-facto contract lives at `docxodus.worker.ts:166` and `:320`, and **no test pins either string**. Reword a worker message and the export silently degrades from the ADR-mandated `resource_limit` code to a generic one — the failure still surfaces (both sites rethrow), but the closed taxonomy erodes exactly where #500 says it must not. Prefer a structured code on the worker response over message matching.

### 2. `verifyOfflineReopen` page lookup is O(n²) — `export-browser.ts:2818`

`expectedPageMap.pages.find(c => c.pageNumber === pageNumber)` runs inside the per-page loop. At the `finalPages` ceiling of 10,000 that is ~50M comparisons during `output_verification`, to resolve records the loop directly above already accesses positionally. Build a `Map` once, or reuse the index.

### 3. Runtime-asset verification is fully sequential — `export-browser.ts:982`

Every asset in the 46-entry closed graph (including the multi-MB `dotnet.wasm` framework files) is fetched, bounded, and hashed one at a time on the `wasm_initialization` path, serializing latency before the worker even starts. `Promise.all` with a small concurrency cap would overlap I/O and hashing.

### 4. Smaller efficiency items

- `sha256` (`:782`) copies its whole input before digesting; `crypto.subtle.digest` already snapshots a `BufferSource`, so this transiently doubles peak memory for inputs up to the 256 MiB HTML ceiling.
- `decodeImages` (`:2208`) awaits each `image.decode()` serially, and runs up to three times per export (two layout attempts plus offline reopen).
- `assertNoClippedContent` (`:2381`) builds a per-descendant `getBoundingClientRect()` array eagerly, then ANDs it with the cheap `scrollHeight` guard that usually makes it moot — test the guard first, as the `.page-content` loop already does.
- `preflightConvertedHtml` (`:741`) re-walks the converted HTML to recount UTF-8 bytes against a limit the worker already enforced with the same algorithm.
- `handleGeneratePackageManifest` (`docxodus.worker.ts:147`) always parses *and* structured-clones both the object and the raw JSON, though each caller uses only one.

### 5. Ripple gap against CLAUDE.md

The new limits-aware `VerificationOps.GeneratePackageManifest(bytes, PackageManifestOptions)` overload reaches WASM and npm, but `tools/python-host/Dispatcher.cs` still calls the option-less overload and MCP is untouched — the ripple checklist marks "new setting or method" for all surfaces. Mitigating: #500 deliberately scopes the export-limits contract to browser/WASM, and MCP never exposed the manifest. Worth a follow-up rather than a fix here.

### 6. Default literals declared twice

`npm/src/export-resource-limits-v1.json` hardcodes `10000 / 1073741824 / 33554432 / 2048 / 1000`, which are precisely the defaults in `Docxodus/Verification/PackageManifest.cs`. CLAUDE.md's "defaults are declared once" rule exists because `undoDepth` drifted this way. Mitigating: #500 defines these as pinned schema-v1 compatibility boundaries, and the export path always passes them explicitly, so the C# defaults never apply there — but the drift risk is real if the C# side moves.

PR #499, opened 2026-08-16
Orchestrate verified delivery bundles

JSv4:
## Summary

Implements the verified delivery-bundle orchestration for #465, stacked on #497.

- derive explicitly named baseline, policy-baseline, working, native-review, and final DOCX states under separate pre-existing/generated revision policies
- prove review accept-to-final and reject-to-policy-baseline identities before publishing proof artifacts
- produce package manifests, stable semantic/package deltas, comprehensive validation, and authoritative #458 receipt evidence through one `DeliveryBundleService`
- inject a capability-declared renderer with exact final/original/markup and comment profiles; preserve renderer fingerprint, page count, warnings, PageMap, and report metadata
- emit a canonical digest-covered manifest with hashes, sizes, MIME types, availability, provenance, and typed artifact relationships
- independently reparse and verify the manifest against separately supplied artifact bytes
- publish only complete bundles to a fresh directory through a verified same-filesystem atomic rename
- expose the shared implementation through .NET, `docxodus-deliver`, and MCP `docxodus_deliver`
- publish the v1 JSON Schema, architecture contract, changelog entry, and an every-artifact E2E gallery

## Verification

- 150/150 delivery, verification, revision-policy, redline-proof, receipt, CLI, and MCP tests pass
- 85/85 full MCP dispatcher regressions pass
- every-artifact E2E requests all 17 schema-v1 artifact kinds, adds two required transaction semantic deltas, publishes a fresh directory, reopens every file, and independently verifies the canonical manifest
- final TRX: `/tmp/docxodus-465/TestResults/delivery-complete-regression/delivery-complete-regression.trx`
- MCP TRX: `/tmp/docxodus-465/TestResults/delivery-complete-regression/mcp-dispatcher-final-regression.trx`
- gallery: `/tmp/docxodus-465-all-artifacts-final/index.html`
- preserved negative-path evidence: `/tmp/docxodus-465/TestResults/delivery-orchestrator/delivery-orchestrator-failed-proof-composition.trx` and `/tmp/docxodus-465/TestResults/delivery-orchestrator/delivery-orchestrator-failed-plan-enumeration.trx`

## Draft boundary

Production standalone paginated HTML/PDF remains gated by epic #434. The core, verifier, atomic publisher, CLI, and MCP surface fail closed when no production adapter is supplied. The every-artifact E2E uses a clearly fingerprinted deterministic test renderer to prove orchestration; it does not promote the LibreOffice evaluation harness or staging HTML as production output.

Depends on #497. Tracks #465.

Issue #440, opened 2026-08-14
[PDF] Validate mixed section sizes and physical page geometry


JSv4:
Parent: #434

## Goal

Ensure each physical PDF page uses the size and orientation of the DOCX section that produced it.

The converter already emits named `@page` rules for mixed-size paginated sections. This issue proves and hardens the complete browser-to-PDF path.

## Acceptance criteria

- [ ] PDF export honors CSS-defined page sizes rather than forcing a global Letter/A4 format.
- [ ] Each PDF page's MediaBox/CropBox matches the corresponding Docxodus page-box dimensions within an explicit tolerance.
- [ ] Automated fixtures cover portrait → landscape → portrait transitions.
- [ ] Automated fixtures cover Letter ↔ A4 or another mixed-size transition.
- [ ] Explicit page breaks, next-page section breaks, continuous sections, and multi-column sections retain correct sequencing.
- [ ] Headers, footers, footnotes, and page-number fields remain associated with the correct section/page.
- [ ] The number of physical PDF pages equals the number of finalized paginator pages.
- [ ] Viewer scaling values do not change physical PDF dimensions or page count.

PR #502, opened 2026-08-16
Add Node PDF export API and CLI

### LINKED ISSUE #439: [PDF] Add a Node conversion API and CLI
Parent: #434

Depends on: the public boundary selected in the export architecture issue and the standalone pagination surface where applicable.

## Goal

Provide a supported Node API and command-line workflow for converting DOCX files to PDF through the paginated renderer and headless Chromium.

## Proposed surfaces

- `convertDocxToPdf(document, options)`
- `docxodus convert input.docx --to pdf --output output.pdf`

The exact package or entry-point location is determined by the export architecture issue.

## Acceptance criteria

- [ ] The API accepts DOCX bytes and returns PDF bytes plus page count, page metadata, warnings, and renderer fingerprint.
- [ ] The CLI accepts input/output paths, review profile, comment mode, timeout, browser executable, and font-directory options.
- [ ] Printing enables backgrounds, honors CSS page sizes, and applies zero browser margins around fixed page boxes.
- [ ] Conversion performs no outbound network requests.
- [ ] Temporary artifacts are isolated and removed after success or failure.
- [ ] Errors distinguish invalid DOCX, conversion failure, browser launch failure, readiness timeout, pagination failure, and PDF write failure.
- [ ] PDF text remains selectable and searchable.
- [ ] Supported hyperlinks and vector content are preserved without rasterizing the entire page.
- [ ] Unit/integration tests cover the library and CLI paths.
- [ ] Installation and usage documentation includes reproducible examples.

JSv4:
Closes #439.

Stacked on #501 / `agent/issue-438-standalone-html` (which is stacked on #500 and the corrected #493). Draft until the export epic stack is reconciled.

## What changed

- adds the separately published `@docxodus/export` Node companion
- exposes immutable byte APIs for PDF, standalone HTML, one-pass HTML/PDF batches, and stable file rendering
- adds the strict `docxodus convert` CLI and a length-framed integration host
- serves only the digest-verified public `docxodus/export-browser` asset graph at an isolated routed HTTPS origin
- launches exact pinned Chromium (or an explicit/caller-owned Chromium), creates a fresh context per render, blocks HTTP(S)/WebSocket/service-worker/download/popup escape paths, and cleans owned resources independently
- prints the reopened finalized HTML with backgrounds, CSS page sizes, scale 1, and zero margins
- parser-verifies PDF header, encryption state, page count, MediaBox/CropBox geometry (0.5 pt), digest, and volatile metadata
- adds stable pre-limited file reads plus fsynced atomic no-replace publication
- preserves structured failed reports across print, verification, and publication failures
- wires companion tests/artifacts into CI and exact-version companion publication into the npm release job

## Verification

- `cd npm && npm run pretest` — pass (core TypeScript, test types, demo logic, 163-file package boundary)
- `cd npm-export && npm test` — 6/6 pass
  - immutable one-pass standalone HTML + searchable/tagged PDF
  - hyperlink annotations and vector chart operators
  - pre-transfer/resource/timeout/browser cleanup failures
  - structured post-layout PDF failure report
  - CLI success, empty stdout, atomic no-overwrite
  - two framed batches sharing one host-owned browser with fresh contexts
- `cd npm-export && npm run test:package-boundary` — pass (39 declared runtime/license files)
- final independent PDF/security audit — no remaining merge-blocking or high-confidence findings

## Viewable artifacts

The local release-candidate gallery is:

`/tmp/docxodus-439/npm-export/test-artifacts/view-artifacts.html`

It links the generated PDF, offline HTML, PageMap, complete report, PDF parser inspection, offline request log/screenshot, hyperlink PDF, vector-chart PDF, CLI output, and structured failure report. CI uploads the same `npm-export/test-artifacts/` tree with `if: always()`.

## Deliberate follow-ups

- explicit font-directory loading remains typed `unsupported_runtime` until #442 supplies the verified pre-layout font runtime
- expanded review/comment profile semantics remain #444
- mixed-section geometry and broader PDF fidelity ratchets remain #440/#443
- framed artifact-ID/delivery diagnostic integration is applied in the pending #499/#465 reconciliation pass


JSv4:
Rebased onto `main` and fixed the red Playwright job.

**Baggage removed.** The PR was showing 20,183 lines across 70 files because it was branched off the #438 branch, carrying stale pre-review copies of the #493, #500, and #501 work. Its genuine payload is **24 files, +6,312/−6** — the `@docxodus/export` companion, CLI, PDF verifier, and framed host. It is now based on `main` with that content byte-identical, and mergeable.

**The Playwright job was already failing before the rebase** (on the original head `be8eee46`), with three failures in `npm-export`:

1. **`fails pre-transfer with stable typed option and resource errors` — a real API bug.** `renderDocxArtifacts`, `convertDocxToPdf`, and `convertDocxToStandaloneHtml` ran `nodeOptionsPreflight`, `snapshotNodeOptions`, and `ownedInput` *before* returning their promise, so an over-limit document threw **synchronously at the call site** instead of rejecting. A caller using `.catch()` would crash, and the test's own `assert.rejects` never saw the error. Fixed by making the three functions `async`: an async body still runs to its first `await` synchronously, so the caller's bytes and options are snapshotted at call time exactly as the export contract requires, while every failure now surfaces as a rejection.

2 & 3. **`CLI writes verified artifacts…` and `framed host rejects local executable authority…` — environment, not code.** Both failed with `browser_launch_failure`. The underlying cause is `Chromium sandboxing failed!`: Ubuntu 24.04 restricts unprivileged user namespaces through AppArmor (`kernel.apparmor_restrict_unprivileged_userns=1`), and the export runtime deliberately keeps `chromiumSandbox: true` per the architecture record ("The owned supported runtime does not opt out of Chromium's process sandbox"). I verified locally that the sandbox is the *sole* blocker — the same executable launches fine with the sandbox disabled — so the fix belongs in CI, not in the product: the workflow now permits unprivileged user namespaces before running the export tests. Weakening `chromiumSandbox` to make tests pass would have been the wrong fix.

Verified locally: after the `async` fix, the only remaining failures on this branch are those two, and they fail solely because this machine has the same AppArmor restriction as the runner.

### One finding left for you

`launchBrowser` attaches the real cause to the error object, but the CLI prints only `"Chromium could not be launched for export."` plus a remediation about shared-library dependencies — which points away from the actual problem. The genuine message (`Chromium sandboxing failed!`) was only recoverable by re-launching Chromium by hand. The architecture record says causes are never *serialized* into reports, but the CLI is a human-facing surface; surfacing the cause on stderr there would have made this a one-minute diagnosis instead of an investigation. Worth considering before the CLI ships.

JSv4:
Red job worked through. Four distinct problems were hiding behind three failing tests:

1. **Sync-throw API bug (product).** `renderDocxArtifacts`, `convertDocxToPdf`, and `convertDocxToStandaloneHtml` validated before returning their promise, so bad input threw at the call site instead of rejecting. Fixed by making them `async`, which preserves the contract-required synchronous snapshot while surfacing failures as rejections.

2. **Contradictory test contracts.** `hardening.test.mjs` asserted `assert.throws` (synchronous) for an aborted signal while `export.test.mjs` asserted `assert.rejects` for the same function — the suites encoded opposite contracts, so no single implementation could satisfy both. Both now expect the rejection the `Promise` signature implies.

3. **Blob frames were JSON-encoded (test).** `framedProtocolRequest` ran blob frames through `JSON.stringify`, so an 11,475-byte DOCX crossed the pipe as a 37,221-byte `{"type":"Buffer","data":[...]}` rendering. The host was right to reject a frame whose length contradicted its declaration — protocol v1 says blob frames carry exact raw bytes and only the control frame is JSON. The forbidden-runtime case masked this because that request is rejected during control-frame validation, before any blob frame is read.

4. **Chromium sandbox unavailable in CI (environment).** Both launch-dependent tests failed with `browser_launch_failure`, whose real cause was `Chromium sandboxing failed!` — Ubuntu 24.04 restricts unprivileged user namespaces via AppArmor. The runtime deliberately keeps `chromiumSandbox: true` per the architecture record, and I verified the sandbox is the *sole* blocker (the same executable launches fine with it disabled), so the workflow now permits user namespaces instead of the product weakening its sandbox.

Local run is **13/15**, and both remaining failures are `browser_launch_failure` from this machine having the same AppArmor restriction as the runner — the CI-side fix is already proven, since the CLI test started passing on the runner as soon as the sysctl step landed.

**A note on diagnosability.** Problem 3 was invisible for two rounds because the test dereferenced `response.batches` directly and died with `Cannot read properties of undefined (reading 'length')`, discarding the host's own explanation. The test now surfaces the fatal envelope, which is how the framing mismatch was identified immediately. The same gap exists in shipping code: `launchBrowser` attaches the true cause to the error object, but the CLI prints only "Chromium could not be launched" plus a remediation pointing at shared-library dependencies — which points away from the actual problem. Worth surfacing the cause on stderr before the CLI ships.

JSv4:
Filed the `launchBrowser` diagnosability gap as #525 so it does not ride along with this PR — the cause is captured at `browser-session.ts:375-383` but `humanError` never reads it, and the printed remediation misdirects toward shared libraries when the real cause is the host's user-namespace policy.

PR #504, opened 2026-08-16
Continue long footnotes within paragraphs

### LINKED ISSUE #489: Footnote continuation stops at the paragraph: an oversized note paragraph is clipped, not split

JSv4:
Closes #489.

## What changed

- Reuses the conservative DOM Range paragraph fragmenter for long footnotes.
- Measures initial and continuation fragments in their exact rendered wrapper context.
- Preserves canonical PageMap anchors and contiguous fragment indices across pages.
- Keeps the bounded fallback for unsafe/indivisible content while allowing later siblings to continue.
- Uploads the Playwright HTML report and raw test evidence on every CI run.

## Root cause and impact

Whole-paragraph continuation left an ordinary footnote paragraph taller than the note band clipped on every attempt. Its invisible tail never received a later page or PageMap fragment. Long text-only footnotes now drain across as many note bands as necessary without duplicating editable DOM identity.

## Validation

- TypeScript typecheck and pagination bundle build
- Focused footnote/PageMap suite: 16/16
- Adjacent pagination regression suite: 25/25
- Final artifact suite: 4/4
- `git diff --check`

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -44,6 +44,12 @@ jobs:
           PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
         run: cd npm-export && npm ci --ignore-scripts
 
+      # The export runtime deliberately keeps Chromium's process sandbox on, so the runner has to
+      # permit unprivileged user namespaces. Ubuntu 24.04 restricts them through AppArmor by
+      # default, which fails every launched-browser export with browser_launch_failure.
+      - name: Permit Chromium's unprivileged user-namespace sandbox
+        run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+
       - name: Install Playwright browsers
         run: cd npm && npx playwright install --with-deps chromium firefox
 
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -70,4 +70,5 @@ jobs:
             npm/playwright-report/
             npm/test-results/
             npm-export/test-artifacts/
+          if-no-files-found: ignore
           retention-days: 7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -461,6 +461,28 @@ All notable changes to this project will be documented in this file.
   unsafe or indivisible layout retains the visible clipped fallback for that one
   element, while later sibling paragraphs continue instead of disappearing inside
   the same clipped band. Documents whose notes fit paginate unchanged.
+
+  Two rules govern the fallbacks. A note is only deferred to the next page when
+  that page's band is larger than what the current page can already offer —
+  otherwise the deferral evacuates the citing page and buys nothing. And the
+  fragmenter's below-line-breaking fallbacks (arbitrary grapheme boundaries, and
+  forcing a first fragment that does not fit) apply only to an element that owns
+  the whole band; with earlier siblings already packed, a paragraph that will not
+  start belongs intact in the next note band rather than cut mid-word. A queue of
+  whole notes larger than the band no longer claims the whole page either: the
+  band takes what fits and the rest of the page still sets body text.
+- **Word's authored footnote separator stories are rendered, and a continued note
+  gets the continuation rule (#489).** `w:separator` and `w:continuationSeparator`
+  were previously dropped, and the paginator drew its own two-inch rule above
+  every note band. Both reserved stories are now converted into the hidden
+  paginated-footnote registry as inert, non-addressable templates; the paginator
+  clones the normal story where a note starts and the continuation story on a
+  page that carries a note over, falling back to typed 2in/full-width rules when
+  a document defines neither. Identities are stripped from these repeated clones
+  so no id, editor anchor, or `PageMap` anchor is duplicated across pages. A
+  marker-only story renders exactly its rule: the pipeline's synthetic
+  empty-paragraph run no longer counts as authored content and no longer draws a
+  blank line under the continuation rule.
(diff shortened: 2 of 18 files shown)


JSv4:
## Rebased onto `main`, now that #440 has landed

This PR previously targeted `agent/issue-440-mixed-page-geometry`, the branch of the closed #503. It is now rebased onto `main`, which carries that work as of #527 (`c69f5ff`).

The rebase was clean: `npm/src/pagination.ts` auto-merged with **no code conflicts**, because `dimensionsFor`, `sectionStartPages` and `adoptEmptyPageOwner` exist again. The only conflict was `CHANGELOG.md`, where `main` still carried the older #488 wording of the footnote entry and this branch revises it into the #489 version — `main`'s other entries are kept and the revision applied.

Two commits were dropped automatically as *already upstream*: the async-entry-point and framed-host-diagnostic changes, which were ports of `main`'s own fixes and are redundant once rebased.

Also removed from this branch's own commits: 47 `SH*.xlsx` test-run byproducts and two `node_modules` symlinks that had been swept in by a `git add -A` (the symlinks slip past `.gitignore`'s trailing-slash pattern). The stack is now 18 files.

### Verification on the rebased stack

- .NET **4262 passed / 0 failed / 3 skipped**
- Playwright **558 passed / 9 skipped**; the only failures are the host's Arial-dependent `tabs-visual` specs, which fail identically on unmodified `main` here

The `standalone-export` clip test that was red on this branch for its whole life now passes, because #527 fixed its cause: a blank running story serialized as a self-closing `<div/>`, which HTML parsers treat as unclosed, nesting `#pagination-container` inside `display:none` staging so the exporter detached the entire page tree during finalization.

PR #527, opened 2026-08-21
Validate and harden mixed-section page geometry (#440)

### LINKED ISSUE #440: [PDF] Validate mixed section sizes and physical page geometry

JSv4:
Closes #440.

Revives the mixed-section page geometry work from the closed PR #503, replayed onto current `main`. The two original commits are preserved as authored; only the conflicts with everything `main` landed since (#500/#501/#502/#524/#526) are resolved here.

## Why this is being re-landed

A document that changes page size partway through — a landscape exhibit inside a portrait brief, or a Letter/A4 transition — is ordinary Word. Without this, the paginator assumes one geometry for the whole document and the exported PDF forces every physical page to that size.

It is also a hard dependency for #504: that PR's footnote packing calls `dimensionsFor`, `sectionStartPages` and `adoptEmptyPageOwner`, none of which exist on `main`.

## What it does

- The shared paginator transfers continuous spill pages to the section that supplies their body while preserving predecessor-owned stories on the shared page, carries section-specific header/footer distances and logical page numbering, selects odd/even stories from each section's one-based page position, inserts story-free blank pages for odd/even section starts, and honors `footnoteLayoutLikeWW8` when a pre-break footnote meets a continuous section.
- PDF verification resolves **inherited** MediaBox/CropBox, `Rotate` and `UserUnit` through the page tree before comparing physical origins and dimensions — quarter-turn rotation swaps physical width/height, and `UserUnit` scales user space, so neither can be read off the high-level page API alone.
- `checkPageCount` admits each physical page before it is allocated, so `finalPages` is enforced without building an over-limit DOM first.

## Conflict resolution

Only three places needed judgement; no code conflicted in the paginator itself.

- `PaginationOptions` — `main` added `deferFragmentIdentities` (#524) and this adds `checkPageCount` at the same spot. Both kept.
- `npm-export/src/pdf.ts` — `main`'s flat `getMediaBox()`/`getCropBox()` reads and its blanket rotation rejection are *superseded*, not dropped: a landscape section legitimately carries rotation, and the boxes are inheritable. Audited line by line to confirm nothing of `main`'s survived only by accident.
- `npm/src/export-browser.ts` — `main` rewrote this file; the only surviving difference is the seven-line `checkPageCount` wiring. Nothing pre-#524 was resurrected.

## Verification

- .NET: **4260 passed / 0 failed / 3 skipped**
- Playwright: **541 passed / 9 skipped**; `pagination-continuous-section`, `page-numbering` and `page-map` all green
- `npm-export` PDF physical-page verification: **3/3** (inherited UserUnit and quarter-turn rotation, non-zero origin rejected, malformed `UserUnit`/`Rotate` rejected)
- TypeScript, test-project and `npm-export` typechecks clean; full WASM build clean

Two local failures are environmental and reproduce identically on unmodified `main` builds here: the five `tabs-visual` specs (Arial absent on this host; Liberation substitutes) and the `standalone-export` indivisible-body-block clip test, which fails the same way with the *base* paginator. CI is the authoritative check for both.


JSv4:
## Known red: `standalone-export` › "fails closed with a report when an indivisible body block would clip"

Everything else is green in CI — the full .NET suite, the whole `npm-export` suite including the six-page Letter/A4 portrait↔landscape PDF fixture that is this issue's central proof, all three build matrices, both Python versions and every CodeQL analysis.

This one test expects `pagination_failure` / `running_story_placement` and gets `output_verification_failure` / `output_verification` ("page 1 has no measurable geometry").

### It is not introduced by the replay

The identical failure is present on `agent/issue-489-footnote-paragraph-continuation`, which is built on the original #503 branch. This defect has been in the mixed-geometry work since it was written; it went unnoticed because #503's Playwright step never ran to completion.

### Safety is intact

The export still fails closed. Nothing incorrect is shipped. What regressed is the *typed contract* — which code and phase a caller observes — not whether bad output escapes.

### Isolation so far

Bisected to the second commit ("Harden mixed-section page geometry"): the test passes with the first commit alone.

Within that commit, ruled out by substituting the first commit's version and re-running:

- `npm/src/pagination.ts` — still fails
- `npm/src/page-geometry.ts` — still fails
- both together — still fails
- the `checkPageCount` wiring in `npm/src/export-browser.ts` — still fails

`npm-export/src/pdf.ts` cannot be involved: this test runs entirely in the browser harness and never reaches PDF verification.

That leaves the `WmlToHtmlConverter` change as the cause. HC006-Test-01.docx has `w:titlePg` and **no** `headerReference` of any kind — exactly the shape this commit changed:

```
// Presence of titlePg selects the First story even when that story is blank.
// Keep an empty registry entry so the paginator does not fall through to the
// Default story on page one.
registry.Add(new XElement(Xhtml.div, ..., "header-first", firstHeaderContent));
```

With `firstHeaderContent` null this now emits an **empty** `header-first` story where `main` emitted none. The converter behaviour looks correct for Word — `titlePg` means page one uses the First story, blank or not, and must not inherit Default — so the suspect is how the paginator sizes a page whose first-page header story is contentless.

Note the plain harness (`test-harness.html`) shows no difference between the two commits for this document, and shows the page paginating normally at 816×1056 with genuine clipping (scroll 1027 vs client 936, descendants overflowing ~90px). The divergence only appears through the standalone export's isolated frame.

### Open question for the fix

Should a contentless first-page header story produce a zero-height band, or should the converter emit a marker the paginator reads without materializing a band? That is a Word-semantics call, so it is left for a decision rather than patched here.

JSv4:
## Resolved: the red `standalone-export` test, at its actual cause

The earlier comment deferred the header/footer gating change as "correct but breaks a contract". That was the wrong call — the gating change was never the bug, it only exposed one. Both it and the geometry work are kept, and the real defect is fixed in `68dc01c`.

### The chain

Selecting a blank First story for a section with `w:titlePg` is right: page one must not fall through to the Default story. But the entry it emits has no content, and a childless `XElement` serializes as `<div .../>`. HTML parsers ignore the slash on a non-void element, so that div is **unclosed** — every following sibling nests inside it, `#pagination-container` included. The container ends up inside `display:none` staging; `finalizePageTree` removes staging during finalization and takes the whole page tree with it.

That is why the symptom appeared three layers from its cause. PageMap measured every page at 0×0:

```
connected=false rect=0x0 inline=612ptx792pt
chain=div.page-box < div.page-container < div.- < div.page-staging
```

so the export reported `output_verification_failure` / "page 1 has no measurable geometry" instead of the `pagination_failure` the test asserts.

### The fix

This is the identical hazard the page-break marker already guards against a few hundred lines above, where the comment calls the empty `XText` child load-bearing (HC008d). That pattern is now applied to all four blank running-story entries — first and even, header and footer — not only the one `HC006-Test-01.docx` happens to reach.

`HC008f` pins it: the entry must render, must not be self-closing for any of the four kinds, and no `data-hf-type` may be left open where the container follows. Verified non-vacuous by removing the four guards and watching it fail.

### Verification

- .NET **4261 passed / 0 failed / 3 skipped** — the gating change and its tests are back, plus HC008f
- Playwright **542 passed**; the clip test passes and only the host's Arial-dependent `tabs-visual` specs fail, as they do on unmodified `main` here

PR #533, opened 2026-08-21
Make the redline reversibility check usable from the browser, npm, Python, and the agent server

JSv4:
## What this is about

When Docxodus generates a redline — a document full of Word tracked changes — you want to be sure it is honest. Two things should be true:

- **Accept every change it introduced**, and you get exactly the document you meant to end up with.
- **Reject every change it introduced**, and you get exactly the document you started from.

And neither of those should quietly consume review markup that was already in the document before Docxodus touched it. If a lawyer had already marked the file up, their edits must survive both operations untouched.

Docxodus can already check all of this. The checker landed some time ago, along with its tests and design notes. The problem is that it could only ever be called from .NET code.

## The gap

The checker had no entry point in `Docxodus/Internal/VerificationOps.cs`, which is the file every non-.NET caller goes through. Searching the browser, Python, and agent-server directories for anything referencing it turned up nothing at all.

So a caller using the npm package, the Python package, or the agent server could *generate* a redline but had no way to ask whether it was faithful. That is the gap this change closes. It adds no new checking logic — the analysis itself is untouched, including its existing handling of moved text, documents edited by several authors, and reporting which part of the package first diverged.

## What changed

One function, `VerificationOps.ProveRedlineReversibility`, becomes the single place that turns a check result into JSON. Everything else calls it, so all callers see byte-identical output:

| How you're calling Docxodus | What you call |
|---|---|
| Browser (WebAssembly) | `DocumentConverter.ProveRedlineReversibility` |
| npm, directly | `proveRedlineReversibility(baseline, intendedFinal, redline)` |
| npm, off the main thread | `worker.proveRedlineReversibility(...)` |
| Python package | `prove_redline_reversibility`, returning typed objects rather than raw dictionaries |
| Agent (MCP) server | `docxodus_track_changes` with `action: "prove_reversibility"` |

## Three judgement calls worth a look

**The two rebuilt documents don't cross the boundary.** Running the check internally produces two documents — the "accept everything" result and the "reject everything" result — alongside the report. Those stay inside the process. Every path listed above is a JSON channel, and base64-encoding two whole Word files into the response would balloon it to carry information the report already summarises as content hashes plus a list of exactly where the documents differ. Anyone who genuinely needs those files can call the checker directly from .NET.

**The browser gets a background-thread option, not just a blocking call.** This check reads three documents and rebuilds two of them, which makes it the most expensive verification Docxodus offers. Offering only a version that freezes the page would be a step backwards from the existing document-validity check, which already has both.

**On the agent server, the document being checked is the session's saved state.** The check needs three documents, but every tool on that server operates on one open editing session. Rather than loosen that rule, the open session *is* the redline, and the two comparison documents are named by file path. Those paths go through the same resolver every other tool uses, so neither can point outside the directory the server was launched against. Checking the saved state — rather than the in-memory working copy with its editing bookkeeping — means an agent verifies the file it would actually hand over. This matches how the existing document-validity check already behaves, needed no new method in the core library, and added no new tool. It is also deliberately excluded from the batching tool, because that tool runs sequences of *edits* and this is a read-only question.

## A smaller change alongside

The browser layer had a size guard that rejected oversized input, hardcoded for at most two documents. It now subtracts each document's size from a running allowance, so a three-document call cannot slip under the 100 MB ceiling through arithmetic that was never written with three in mind.

## How it was checked

New tests on three surfaces, all confirming the same things: identical output every run (a saved report's hash depends on it), content hashes that match the exact inputs passed in, and safe behaviour on bad input — when a file can't be read as a valid Word document, neither the accept path nor the reject path is attempted at all, so a half-finished result can never be mistaken for a clean verdict.

- `Docxodus.Tests/RedlineReversibilityTransportTests.cs` also covers the agent-server path end to end, including refusing a file path outside the permitted directory, and confirms the new action is offered but not batchable.
- `python/tests/test_redline_reversibility.py` checks that the report decodes into real typed objects, down to nested values, and that each path names the document it is expected to reproduce.
- `npm/tests/redline-reversibility.spec.ts` drives the browser build through the test harness.

All 14 CI checks pass, including the full .NET suite, the browser build, and both Python versions.

Worth knowing: the machine this was written on has no .NET SDK, so none of the C# was compiled locally. It was written against interfaces read directly from the codebase, and CI is what verified it. It built and passed first time.

## The second commit

`chore(git): renormalize Docxodus.Tests.csproj line endings` is unrelated housekeeping that also appears on #532. That file's stored copy has Windows line endings, which contradicts the rule in `.gitattributes`, so git reports it as modified in every fresh checkout. The change is line endings only. Whichever of the two lands first, the other becomes a no-op.

Relates to issue #464.


JSv4:
## Review record

Reviewed against issue #464 (facade correctness, per-transport wiring, test non-vacuity, succinctness). Verdict: the facade and all five transport wirings are sound — every surface routes through `VerificationOps.ProveRedlineReversibility`, the npm/python wire types were verified field-for-field against the C# canonical serializer, the WASM size-guard rework is arithmetically identical for the existing two-document callers, and the MCP action resolves both paths through the same symlink-resolving containment check `docxodus_open` uses, with batch exclusion doubly enforced.

Two review findings were the bypassable kind, fixed in the follow-up commit:

- **The npm worker path had zero coverage** while the spec header claimed it: the spec only drove the raw WASM export. A misspelled worker `case` or a dropped `proof` field would have stayed green. The spec now drives `worker.proveRedlineReversibility` end to end (compare → prove → prove again) with the same schema/digest/direction bindings as the direct-export test.
- **The MCP path-escape test refused a file that didn't exist**, so a plain read failure produced the same exception as the containment check — deleting the resolver's scope check would not have failed the test. It now writes real package bytes outside the store scope and asserts the "outside this server's document scope" message.

Also: a behavioral batch-refusal test (RT005 submits a `prove_reversibility` step to `docxodus_mutations` and asserts the whitelist rejects it, rather than only grepping schema text), two duplicated TS types now alias the session registry's existing ones, and the python docstring no longer implies cross-surface string equality that the camelCase-vs-snake_case wires make false.

One suggestion deliberately **not** taken: asserting `success == true` on happy paths. Probed empirically — a generated redline over independently-built inputs reports `success: false` by design (normalized whole-package divergence; same family of reasons as the engine's own RP001), so that assertion would be wrong, and the digest/direction bindings are the correct transport-seam pinning.

Local verification: RT001–RT005 5/5, full reversibility filter 188/188, `npm run typecheck` clean, all 3 Playwright specs (including the new worker one) green on a fresh local build.

Issue #466, opened 2026-08-14
[Eval] Build a realistic document workflow evaluation suite


JSv4:
Parent: #436

Related to: #434 and #435

## Goal

Measure whether automated document workflows complete intended legal-document tasks precisely, preserve unrelated content, and produce valid reviewable artifacts.

## Initial scenarios

- Replace a defined commercial term at the correct occurrence(s).
- Amend a notice period with native tracked changes.
- Insert a clause while preserving styles and numbering.
- Modify economics in a table.
- Add a comment, reply, footnote, and internal cross-reference.
- Preserve signature blocks, headers/footers, fields, and section layout.
- Fill content controls in a template.
- Compare and consolidate multiple document versions.
- Operate on a document containing pre-existing revisions/comments.

## Acceptance criteria

- [ ] Every scenario contains a tracked input fixture, machine-readable instruction, expected outputs, and deterministic invariants.
- [ ] Metrics include task completion, target precision, unintended anchor/part changes, document validity, redline reversibility, and rendering regression.
- [ ] Evaluation separates engine/tool correctness from model planning quality.
- [ ] A deterministic scripted caller establishes the engine baseline before agent runs are scored.
- [ ] Failures preserve inspectable DOCX, HTML/PDF, semantic diff, receipt, and visual artifacts.
- [ ] Corpus provenance and redistribution permissions are recorded.
- [ ] CI has a fast deterministic subset; the full corpus is available as an opt-in or scheduled gate.
- [ ] Adding a new scenario requires explicit expected invariants rather than visual inspection alone.

PR #534, opened 2026-08-21
Add a deterministic test suite that measures whether document edits land precisely

JSv4:
## What this is about

There is a difference between "the edit worked" and "the edit worked and nothing else moved". Ordinary tests catch the first. This adds a suite that measures the second.

Each case describes a realistic document task — rename a defined term in one clause, change a notice period as a tracked change, adjust one figure in a fee table — and then scores the result on several questions at once:

- Did the intended change actually land?
- Did *only* that change land, or did something nearby get caught too?
- Did unrelated parts of the file survive intact?
- Is the resulting document still structurally valid?
- For a tracked-changes task, can the redline still be cleanly accepted and rejected?
- Does the document still render?

The scoring reuses machinery Docxodus already has rather than inventing new comparisons: an existing structured record of what changed between two versions supplies the precision count, existing package inventories supply the collateral count, the existing document-validity gate supplies validity, the existing redline check supplies reversibility, and the HTML renderer supplies the rendering check.

## Anatomy of a case

One JSON file under `eval/scenarios`, validated against a schema that ships alongside it:

| Field | Answers |
|---|---|
| `fixture` | What document did we start from? |
| `intent` | What would a person have asked for, in words? |
| `steps` | Which tool calls perform the task? |
| `invariants` | What makes this a pass — and what makes it a *precise* pass? |

## Three design decisions worth a look

**The runner calls the agent tool surface, not the .NET API.** This suite is meant to establish how well the engine itself performs, and the engine has to be measured where an agent actually meets it. Driving the internal .NET methods would hide problems that only exist in the tool layer — a missing argument, a guard that fires wrongly, an action that silently does nothing.

The `intent` field is deliberately *never read* by the runner, which executes the listed steps verbatim. That separation is the point: because no planning happens, a failure here is always the engine's and can never be blamed on a model choosing badly. Later work that does score an agent's planning will reuse these same documents and the same pass criteria, measured against this baseline.

**Fixtures are build scripts, not committed `.docx` files.** Each one is a list of steps replayed over a blank document. Two consequences: the suite contains no third-party document content, so there is no licensing question to answer (recorded in `eval/PROVENANCE.md`), and a fixture can be reviewed as a readable diff instead of an opaque binary.

One test asserts each fixture builds twice to the same content and the same file inventory — deliberately *not* the same bytes or the same hash, because internal identifiers are generated fresh on every build. Two honest builds differ at the hash level while being the same document; asserting otherwise would be asserting that bookkeeping is stable, which it is not and does not need to be.

**Steps find their target by content.** A step says "the paragraph containing this phrase" rather than reusing whatever the previous step returned. That keeps each step independent of the previous one's response format, and it makes drift fail loudly: if a phrase matches fewer places than expected, the run stops with an error instead of quietly editing a different paragraph and then reporting excellent precision.

## What ships

- **Renaming a defined term** — the sample agreement uses "Service Provider" in three places; the task renames it in one, and the pass criteria explicitly name the other two as content that must survive.
- **Amending a notice period** — thirty days becomes sixty, recorded as real Word tracked changes, then put through the redline check.
- **Editing a fee table** — one year's figure changes, between two adjacent rows that must not move.

A second test refuses any case that declares nothing measurable. It requires a statement of what had to change, a limit on how much was allowed to change, and a non-empty list of what had to survive — so "it looked right" cannot serve as a pass.

## What the first CI run changed

The suite compiled and ran on the first attempt. Four of **its own** assertions turned out to be wrong, and the third commit fixes them. All three causes were in the suite; none were in the engine. Two changed the design, so they are worth recording:

1. **Text checks were searching a Markdown rendering of the document.** That rendering represents a table structurally, so cell text does not appear in it as ordinary prose — which made a *correct* fee-table edit score as a total loss. The giveaway was that the same run's rendering check for the new figure passed while the text check for it failed. Text checks now query the document through its own search, which is also how the steps locate their targets.

2. **The "what should this become" document was produced by opening a second editing session.** That session stamped its own settings and bookkeeping onto the file, and the redline check correctly reported that bookkeeping as a discrepancy belonging to the redline. It is now produced by the library's own whole-document accept operation, which leaves no such trace.

3. **The reproducibility test compared file hashes**, which as described above asserts something that was never true.

**One check is deliberately switched off, and it is the change I would most want a second opinion on.** Requiring the two documents to match *exactly* would really be asserting that my way of producing the expected document agrees byte-for-byte with the checker's own internal path — a claim about two pieces of engine machinery rather than about whether the redline is reversible. The engine's own existing tests expect that exact match to fail on a generated redline for the same family of reasons. What is still always required: both directions complete, and pre-existing review markup survives. Tightening this properly needs a fixture whose expected result is stated up front rather than derived, which is follow-up work noted in `eval/README.md`.

Everything else held on the first run: all three precision limits, every collateral count, and every validity verdict.

## Diagnosing a failure

A failing case writes out both documents, the rendered HTML, the extracted text, the structured record of what changed, and the redline report when one was produced — so a failure can be understood without re-running it. Set `DOCXODUS_EVAL_ARTIFACTS` to choose where they land.

## Scope

This is the fast, deterministic portion: no PDF, no browser, no network, so it runs on every push. Deliberately left out, and tracked separately: PDF and visual-difference scoring (that belongs with the separate PDF fidelity work, and depending on it before it exists would bake unstable numbers into these baselines), scoring of agent planning, and the remaining task types — inserting a clause with numbering, authoring comments and footnotes and cross-references, filling template fields, merging several versions, and working on documents that already carry someone else's tracked changes.

## How it was checked

The full .NET suite passes on the current head, so all three cases meet their criteria. The case and fixture JSON validate against the schema. All 14 CI checks pass.

Worth knowing: the machine this was written on has no .NET SDK, so none of the C# was compiled locally — it was written against interfaces read directly from the codebase, and CI is what verified it.

## The second commit

`chore(git): renormalize Docxodus.Tests.csproj line endings` is unrelated housekeeping that also appears on #532 and #533 — that file's stored copy has Windows line endings, contradicting the rule in `.gitattributes`, so git reports it as modified in every fresh checkout. Line endings only; whichever lands first, the others become no-ops.

Relates to issue #466.


JSv4:
## Review record

Reviewed against issue #466 (harness correctness, invariant non-vacuity, scenario/fixture quality, succinctness, CI cost). Verdict: the harness genuinely drives the MCP dispatcher — every consumed tool contract and response shape was verified against main — the scoring directions are correct (no swapped accept/reject, part add/remove not inverted), every invariant needle matches the fixture verbatim, and no catch block converts failure into a pass. The 1511 lines are proportionate: ~700 harness+tests, ~450 schema+scenarios, ~200 fixture, ~160 docs, with almost every schema knob exercised.

Three review findings sat in exactly the vacuous-assertion class this suite exists to prevent, fixed in the follow-up commit:

- **Reversibility asserted completion, not restoration, and pre-existing preservation was unfalsifiable** (the fixture carried no pre-existing revisions, so the preservation flag held over an empty set). The fixture now ends with one live tracked change by "Prior Reviewer" — probed: the proof classifies 2 pre-existing revisions alongside 2 generated — and a new default-on `rejectMustRestoreBaseline` invariant asserts the reject path is *semantically equivalent to the opening package*. The reject path is the honest one to assert: its expected document is stated up front, never derived. (Probed on the accept path too: the derived intended final now demonstrably diverges semantically, which is precisely the derivation-scoring problem the README describes — confirming `mustSucceed` staying opt-in is the right call.)
- **EV002's scoreability gate had holes**: an empty `targetPrecision` passed, and a typo'd invariant key was silently ignored by every checker despite the schema's `additionalProperties: false` (which nothing enforced). EV002 now requires at least one precision bound and rejects any invariant key outside the checkers' vocabulary; the schema now requires `reversibility.mode` and `minProperties: 1` on `targetPrecision`; the README no longer claims schema validation that nothing performed.
- **Step results were discarded**: edit tools report failure as `success: false` rather than throwing, so a failed step would have continued the run and surfaced as a misattributed invariant failure — or, for a needle already present in the fixture, not at all. The harness now stops on the first failed step.

Also: EV003 gains a sentinel-phrase anchor (identical-but-empty projections can no longer count as reproducible), failure artifacts now include the full verification report, and the duplicated part-inventory helper and a dead response-shape fallback are gone.

Local verification: full .NET suite 4270 passed / 0 failed on this head; the eval subset (18 tests) runs in under a second, keeping the fast-deterministic claim honest.

Remaining scope vs #466 is honestly stated in `eval/README.md`: 3 of 9 scenario families ship here; clause insertion, notes/comments/cross-references, content controls, N-way consolidation, and agent-scored runs are follow-up.

Issue #614, opened 2026-08-28
insertFootnote/insertEndnote under render_inline recording write an irreversible mark instead of refusing


JSv4:
> **Correction (2026-08-31, after the fix in #625).** The mechanism described below was wrong in
> one respect, and the correction is worth reading before this issue is used as a reference: the
> body citation was **not** wrapped in `w:ins`. `InsertFootnote`/`InsertEndnote` never consulted
> `TrackedChangeMode` at all, so the citation *and* the definition were both written as ordinary
> untracked content and reject-all had nothing to reject. That is why the reproduction below
> reports **two** residual revisions rather than one. The symptom, the severity and the proposed
> resolution were all correct; only the encoding was misread. The corrected paragraph is inline
> below.

## Summary

`DocxSession.insertFootnote` / `insertEndnote` applied while the session records mutations as tracked changes (`TrackedChangeMode.RenderInline`) produce a mark that **reject-all does not undo**, and — unlike the neighbouring operations — they do not refuse.

~~The reference in the body is wrapped in `w:ins` and is removed on rejection, but the note *definition* written into `/word/footnotes.xml` (or `/word/endnotes.xml`) stays behind.~~

**Corrected:** neither is recorded. The op ignores the recording mode entirely — the body citation is written as a plain `w:r` carrying the `w:footnoteReference`, with no `w:ins` anywhere, and the definition goes into `/word/footnotes.xml` (or `/word/endnotes.xml`) as ordinary content. A reviewer therefore has nothing to accept *or* reject, and reject-all is a no-op. The rejected document does not reproduce the baseline, and the redline never showed the insertion as a change in the first place.

## Why this is a defect and not a documented limitation

The comparable operations **fail closed**. `ApplyListFormatRange` under the same recording mode refuses with:

```
tracked_operation_unsupported: ApplyListFormatRange has no reversible native
tracked-change encoding on this document shape; no changes were made
```

That is the right behaviour: nothing is written, so nothing can fail to be taken back. Note insertion silently writes something it cannot take back instead. Of the two failure modes this is the worse one, because the redline looks complete and only fails when somebody actually rejects it — potentially long after the document has left the building.

## Reproduction

Against a baseline document, with a session recording in `render_inline`:

```js
const session = openDocxSession(baselineBytes, { trackedChanges: TrackedChangeMode.RenderInline });
const target = session.findAllByText('some clause')[0];
session.insertFootnote(target.id, 0, 'Negotiated on March 11; see side letter.');
[...]

PR #615, opened 2026-08-28
Add REDLINE THEATER: an MCP-driven redline demo that proves itself

JSv4:
A third demo page at `docs/demo/redline.html`, with two modes. The Arcade painted frames into one paragraph through `raw.replaceXml`. Golf made the editing surface the game and the comparison engine the referee. This one puts the agent protocol on screen — and then measures what the comparison engine actually costs.

## Mode 1 — the negotiation

REDLINE THEATER

Three counsel negotiate a Master Services Agreement. Every edit that lands is dispatched from a real JSON-RPC 2.0 `tools/call` frame — the shape `docxodus-mcp` accepts over stdio — streamed on a wire console beside the document with its measured duration.

Nothing in this mode renders a diff. The session records in `render_inline` mode, so each call writes native `w:ins`/`w:del` into the live package and the editor repaints only the block that changed. What you watch is the file that downloads. Because the three counsel are three values of the session's revision author, the markup is genuinely per-reviewer — the panel's breakdown is read back out of `listRevisions()`, not tracked separately.

You can see the negotiation layered in the markup: Net 60 struck to Net 30 struck to Net 45; the liability cap struck from trailing fees to 2x to 1.5x. That history is the document, not an annotation on it.

**The finale proves the result rather than asserting it.** `proveRedlineReversibility` rebuilds two packages from the redline; `docxDiffGetRevisions` independently answers the content question. Accept-all reaches the negotiated final, and reject-all restores the baseline with **zero content differences**. Reject-path package divergences are classified against a closed, written-out set of parts that review comments legitimately explain (`comments.xml` and the content-types, rels, styles and settings entries that come with it — rejecting cannot remove a comment, because a comment is not a tracked change). Anything outside that set fails the verdict, which is how the note-insertion defect below was caught.

**One call is expected to be refused.** Converting the carve-outs to a Word-native `(a)(b)(c)` list means writing `w:numPr`, which has no reversible tracked-change encoding — so the engine declines rather than writing a mark reject-all could not undo. The wire shows that in amber. An engine that fails closed is the same guarantee the proof verifies, enforced up front.

## Mode 2 — diff stress

The first draft of this PR called itself REDLINE THEATER while using the comparison engine exactly once, as a verification cross-check; `docxDiffCompare` was never called at all. That is a fair thing to be pulled up on — "redline" means two things in this codebase and only one was on screen.

DIFF STRESS

So the page now has a second mode that does the other thing: recompute the redline from scratch after every single edit, baseline against current, and time it. Each frame appends a clause, so the input grows under the engine rather than sitting at one fixed size that would measure nothing.

Three selectable depths, because "how fast is the diff engine" has three honest answers. Measured in-browser on the demo's own agreement, each row a median of three full 40-frame runs:

| depth | engine calls | rate | p50 | vs. recording |
|---|---|---|---|---|
| `revisions` | `docxDiffGetRevisions` | ~19/s | 52 ms | 33× |
| `redline` | `docxDiffCompareProducts` → package + revisions | ~13/s | 74 ms | 45× |
| `full + HTML` | the above, then `convertDocxToHtml` | ~7/s | 141 ms | 78× |

A mutation through the same MCP endpoint costs ~2 ms. So a redline-per-edit loop lives between roughly **7 and 19 frames per second** on a 3 KB document. "Can we animate a diff per edit at 60fps" is still no — but it is a much closer no than when this PR opened, where the same loop ran at 2 to 6. The gap is still the whole argument for recording a redline when you are the one editing and computing one when somebody hands you a document they changed: both emit the same native markup, and the meter shows what each costs to get there. The panel reports the ratio it just measured, not the numbers in this table.

The `redline` depth uses `docxDiffCompareProducts` rather than `docxDiffCompare` followed by `docxDiffGetRevisions`, because one memoized alignment pass yielding both products measured **51 ms against 85 ms** for the two calls separately — about a third saved, which is what that API is for.

**These numbers have moved five times, which is the point** — this branch has merged `main` repeatedly while sitting green, and every figure shifted without a line of the panel changing. Four movements have a cause; one does not, and separating those is what the list is for:

- **#616** removed a read amplification; `docxDiffCompare` 278 → 153 ms.
- **#626** gated the compatibility normalizer on a streaming scan. The two deeper rows came down 10–18% while `revisions` held inside the noise band — a saving that grows with pipeline depth, which is what a load-path change looks like.
- **#627** stopped giving the engine's reads an identity nothing asks for; all three rows moved together, 13–18%. Uniform across depths is the signature of a change on the path every depth shares.
- **A refresh with no attribution.** After #620's `WmlToHtmlConverter` speedup every row read 5–15% *slower*, which nothing in that merge could do — `revisions` never calls the converter. Decomposing said the opposite: the conversion stage alone went 154 → 143 ms, cheaper, exactly as #620 intends. That reading was the machine on a different day, and it is labelled as such rather than pinned on the nearest PR.
- **#653 compiles the browser build's hot paths ahead of time from a recorded profile, and everything roughly halved.** `revisions` 143 → 52 ms, `redline` 170 → 74, `full + HTML` 324 → 141; `docxDiffCompareProducts` on fixed inputs 125 → 51, and the conversion stage in isolation 144 → 60. Uniform across depths again, but the shared thing is the .NET execution underneath them rather than a code path — which is why the ratio against the recording path *moved* (68–157× → 33–78×) instead of holding: recording got faster too. A claimed 2× got more evidence than a single session, both methods agreeing and the controlled figures pooled across two.

Those attributions are what the *shape* of each movement suggests; the panel measures the total and the commit messages are the authority on cause. Where the shape fits nothing in the diff, the honest label is "no cause found". (#629 is a standing example of the same discipline: it landed with #627 but its snapshot reuse is *across* comparisons, and this loop makes one comparison per frame, so it is not credited.)

`benchmarks/docxdiff-stress/FINDINGS.md` asks this question far more rigorously — 147 KB certificate of incorporation, stage attribution, allocation figures, medians of nine runs. That harness is the authority on engine performance; this mode is the watchable version on a small document, and the README says so.

**A measurement trap worth flagging to a reviewer**, since the branch fell into it once: the stress p50 is taken while the document is *growing* under the loop, so it depends on how far the run got, and one container spreads 5–20% on fixed inputs anyway. It is therefore unfit for comparing engine builds, however much it looks like a benchmark. After #623 the stress readout appeared to show a third-sized regression while a controlled fixed-input measurement showed none — the controlled one was right. Both `diff-stress.js` and the README say so, and every "did this move?" question above was answered the controlled way (fixed baseline/current, medians of nine, three reps) before the table was touched.

Other details worth knowing: the meter resets per run (carrying frames across a depth change reports a p50 describing neither pipeline — the first draft did exactly that and reported an identical 206 ms for all three depths, which is how the bug was found); fps comes from the median frame rather than the mean, so one hitch does not halve an otherwise-steady reading; leaving the stress pane stops the loop so it cannot burn the engine behind hidden UI; and the appended clauses are distinct real sentences, since repeating one paragraph would be an unrealistically easy alignment problem that flatters the numbers.

## How the frames can honestly be called real

`tools/mcp-server` is a .NET process on stdio, and a browser cannot subprocess it. So `docs/demo/mcp-wire.js` reimplements the server's **front half** — envelope parsing, the (tool, action) routing table, the `content[].text` + `isError` result shape, and the convention that a business-level failure is a tool *result* rather than a JSON-RPC error — over `DocxSession` via WASM instead of over a pipe. The frames are the frames; the transport under them differs.

That correspondence is checked rather than claimed. `docs/demo/tools/redline-theater.test.mjs` parses the shipped `ToolCatalog.cs` and fails if any (tool, action) pair, search `mode`, `set_mode` value, or argument the script puts on the wire is absent from the real catalog. It earned its keep immediately, catching two mismatches:

- a batch argument I had called `atomic`, which the catalog calls `mode` with an enum;
- `set_mode` wire names (`accept`, `render_inline`) passed straight to `setTrackedChanges`, which takes the `TrackedChangeMode` enum — the mode silently did not switch, so the "clean" baseline was being authored with recording on.

Deliberately not reimplemented: the document store and its containment checks (a tab has no filesystem, so sessions are opened by the host page rather than by `docxodus_open` with a path), the transaction/retry journal, the delivery-bundle host, and the MCP Apps UI extension.

## A defect found while building this, and now fixed upstream

`insertFootnote`/`insertEndnote` under `render_inline` recording wrote the citation *and* the note definition as ordinary content, so reject-all removed neither: the note's text survived a "full" rejection and the redline was not reversible. Unlike `ApplyListFormatRange` under the same mode, they did not refuse either. The inconsistency was the defect more than the irreversibility — a silent irreversible mark is worse than a loud refusal, because the redline looks complete until somebody rejects it.

Filed as #614; fixed by **#625**, then completed by **#638**. #625 made the citation the reversible unit but left the definition unmarked, compensating with an unconditional prune on the stateless reject — which was ambiguous, because a `w:ins` citation beside an *unmarked* definition is also exactly what a comparison emits when the counterpart merely cites a husk the baseline already owned, and that prune ate the husk. #638 marks the definition's content too and guards the prune on it being emptied as well as orphaned.

Act II footnotes the negotiated cap as a result, so the proof exercises that path on every page load. The guard test that had pinned the paragraph workaround is inverted rather than deleted — the script must still contain exactly one note-insertion beat — and the browser assertion checks **both** halves: the citation run inside a `w:ins`, and the definition Act II authored carrying insertion markup, identified by its text because the part also holds Word's two reserved separator notes, which carry no markup and must not. The `tracked_changes.md` "Known gap" section is retired, pointing at the encoding docs #625 and #638 added.

Worth noting for the reviewer: the marker goes at the **end** of the clause rather than against the "1.5x" figure it qualifies, and that is the engine's constraint rather than a style choice. That figure is itself a tracked insertion, and citing into it would nest a reference run inside a `w:ins` — a revision with no independently meaningful resolution. `InsertFootnote` refuses that outright, which is the right answer. The script therefore searches the clause's untouched tail and cites past it, using a `bindOffset` step field so the offset is looked up rather than hardcoded — a literal would drift the moment any earlier act changed the sentence.

## Two things fixed in my own diff

- **Repaint on mutations, not on every successful call.** The endpoint fired its repaint callback after every non-erroring call, so a search that changed nothing still scheduled a repaint — 29 refreshes for 30 calls. Worse, the spec's "repaints coalesce" assertion was passing 29 &lt; 30 *only* because the one deliberately-refused call does not fire the callback: it claimed to measure coalescing while actually measuring the refusal. `resultMutated()` now decides, and it is 13 repaints for 31 calls.
- **The GIF capture script could not run the way its own header documented.** It lives in `docs/demo/tools/` but runs from `npm/`, so a bare `import '@playwright/test'` resolved against the wrong directory. It now resolves from the working directory.

## One observation about the tool surface

Attribution is a session setting (`revisionAuthor`, taken at `docxodus_open`), so three counsel in one session is a host-side switch with no tool behind it. The demo does it off-wire and the README says so. Whether the MCP surface should be able to express "these next edits are by a different reviewer" is a real question, but not one this PR answers.

## Testing

- **14 browser assertions** (`npm/tests/demo-redline.spec.ts`): ten for the negotiation (clean baseline, full run, per-counsel attribution, the reversibility proof on screen and independently, the footnote being a real tracked citation with both halves marked, the accepted end state clause by clause, the saved package carrying `w:ins`/`w:del`, the wire, a latency budget, a second clean run after reset) and four for the stress mode (a real diff every frame against a growing document, a deeper pipeline measurably costing more — the assertion that catches the depth selector silently not applying — the recorded-vs-computed ratio derived from both halves, and leaving the pane stopping the loop).
- **61 headless checks** across `docs/demo/tools/redline-theater.test.mjs` (45: frame shapes, bindings, script validation, telemetry, verdict logic, the catalog contract) and `diff-stress.test.mjs` (16: the meter and its reset, sparkline geometry including flat and single-point cases, the budget classifier, both ratios, and that no two appended clauses repeat).
- Re-verified after every `main` merge against a rebuilt WASM engine — including the one that removed the `WmlComparer` engine entirely (#643, ~20k lines), where every call this demo makes was checked to still exist before trusting a green run.
- Conflicts resolved along the way: `CHANGELOG.md` four times (both sides adding under `[Unreleased]`, both kept), `npm/package.json` once (both branches extending `pretest` and `test:demo-logic`, taken as a union and verified file by file), and `docs/demo/README.md` once (the v11.0.0 pin *and* the page count — resolved to seven pages at the current pin, which is both sides). The v11.0.0 merge also needed a manual fix: git's auto-merge swept this demo's changelog entry out of `[Unreleased]` into the released section, which would have claimed it shipped in a version it is not in.
- `npm run test:package-boundary` confirms none of this ships in the npm tarball.

A note for anyone editing the script: `## 8. Governing Law` loses its `8.` to the markdown ordered-list parser once the `## ` is stripped, so section headings are written as `Section 8 — Governing Law`. There is a test for that too.

Both GIFs are regenerable rather than mystery binaries — `docs/demo/tools/capture-redline-gif.mjs` (`MODE=stress` for the second) and `frames-to-gif.py`, neither run by CI. They play at 1.02× real speed; for a performance demo that matters. Both were recut after #653 halved the engine, because a stale meter on a performance demo is misleading rather than merely old. The stress GIF's meter reads higher than the table above — 132 ms against 74 — and that gap is real rather than an error in either: capture runs a 10fps screenshot recorder against the same CPU, and the clip starts after a full negotiation has already grown the document. The README says so at the GIF, so a reader meeting the two numbers does not conclude one of them is lying.

https://claude.ai/code/session_01McoXsjiXqjyRh17URzDDEU


JSv4:
CI note: the `test` job is red on `fb193b8`, and it is not this PR's failure.

The failing test is `npm/tests/docxodus.spec.ts:480` — "CZ Comparison Tests (tracked changes) › move detection exposes MoveGroupId and IsMoveSource in revisions" — which times out at 60 s inside `compareDocuments` on all three attempts. This PR's diff is 16 files of documentation, demo JavaScript, one Playwright spec and CHANGELOG; it contains no C# and cannot reach the comparison engine.

Bisected rather than assumed. Building and running the same single test on each side of the merge:

- `5339d04` (#691, immediately before #693): **passes in 2.7 s**
- `4652f65` (#693) and this branch's merge of it: **60 s timeout**

So #693 took that comparison from 2.7 s to over 60 s. Filed with the bisect and a pointer at the likely cause as #695 — #693's own `perf(diff)` commit documents the same failure mode on a different pair ("49 seconds where the comparison had taken 1.3") and adds a prefilter that this pair appears to slip past.

I have not ported a fix because none exists yet, and the fix belongs in the diff engine's alignment logic rather than in a demo PR. I am also not re-running the check: it reproduces deterministically on three attempts in CI and again locally, so a re-run would burn ~20 minutes to fail identically.

This PR's own suite is unaffected — 14/14 browser assertions and 61/61 headless checks pass on the merged build. I will keep it watched and merge `main` again once #695 is fixed.

---

PR #620, opened 2026-08-29
Run the actual Doom engine as a fourth cartridge, keeping the E1M1 raycaster

JSv4:
THE DOCX ARCADE's third cartridge was a hand-written ASCII raycaster walking Freedoom's E1M1, rasterized offline into a character grid — a real Doom *level* in a Word document. The arcade now also holds a real Doom *engine* in a Word document, as its **fourth cartridge and its default**: id Software's own code, by way of [doomgeneric](https://github.com/grubbyplaya/doomgenericjs) compiled to JavaScript, running on Freedoom's IWAD, with every frame becoming the media payload of one inline image in a Word paragraph, swapped through `DocxSession` and re-rendered by the editor's incremental block render. The rasterized E1M1 cartridge **stays beside it as cartridge 3** (restored by #639, merged into this branch — see the follow-up section at the end): the two are different demos, and the reason they both earn a slot is the heart of this PR.

Real BSP rendering, real monsters, doors, weapons, menus and status bar — and it is still only a document. **Esc** freezes the inline image as ordinary content: **Ctrl+C / Ctrl+V** duplicates the frame as a second native document image, **Undo / Redo** becomes a frame scrubber, and Save writes it to `.docx`.

The playable frame is Doom's complete, lossless **320 × 200 colour framebuffer** — view, weapon, ammo, health, armor and the status face — at or above a ten-FPS design point. Getting there took two designs that did not work, and most of this description is about why.

## Why an inline image, and not the character grid

The obvious way to put a game in a Word paragraph is to make the paragraph a screen: a grid of characters, each cell an ink and a `w:shd` shading. That is how the arcade's text cartridges work, and it is what this cartridge did for most of its life. The cost model that governs it is worth keeping, because it is a real property of the format:

> A run breaks when the **ink** or the **`w:shd` shading** changes. It never breaks on the **glyph**, because every character in a span shares one `w:t`.

So colour is scarce and glyph variation is free, one OOXML→HTML conversion of the paragraph is linear in runs, and the whole exercise becomes an allocation problem. Several honest answers came out of that, and they are still in the tree for the text cartridges: quadrant block glyphs carrying four sub-pixels per cell for no extra runs, a budget-driven merge that squeezes a frame to a fixed allowance by repeatedly merging whichever two adjacent runs lose the least picture, and block-truncation endpoints so a run's two colours are the group means either side of the span's mean luminance rather than its outliers.

**None of it survives contact with the thing a player actually needs.** Character projections met the run budget only by destroying the evidence: ammo, health and armor numerals. Adding cells restored the fidelity and pushed the standard converter below 10 fps — 5pt cells sampling 304 × 100 were sharper in the abstract and illegible at the size anyone views the demo. The run floor is unforgiving here: a run cannot cross a line break, so a frame can never use fewer runs than it has picture rows, and resolution bought in rows is bought at the worst possible price.

A native `w:drawing` is both more honest and faster. `replaceImage` updates the actual package media, `editor.refresh()` re-renders that one changed block, and Save, Undo and reopen all see the same frame. At 451 × 338.25 points the 320 × 200 image is corrected to Doom's 4:3 display aspect and the 11-pixel HUD digits land about 25 CSS pixels tall. **The browser never mounts an out-of-document game canvas** — that is the whole claim, and the image path keeps it while the grid could not.

The loading and error states are still a text grid, and the framebuffer painters remain exported as test helpers, exercised by the headless canvas-font guard. The playable path does not call them.

## The library changes this needed

**Incremental block render now carries embedded images.** A single-block re-render clones the target into a reusable throwaway shell. OOXML relationship ids are scoped to their owning package part, so cloning only the XML leaves a perfectly valid-looking image pointing at no part in the shell — which `WmlToHtmlConverter` correctly omits, and the block renders without its picture. The renderer now copies the referenced image parts into the shell's main part and rewrites the clone's relationship ids, and drops the prior render's copies first, since the shell owns only formatting parts between renders. Without this the whole approach is impossible: every frame would render blank.

**Preparing a document for conversion got cheaper.** Profiling one conversion put 45% of the time in `MarkupSimplifier` and 25% in `FormattingAssembler`. The simplifier's most expensive step was the rule deciding which adjacent runs may merge: it built a STRING per run (the run's kind plus its run properties through `XElement.ToString`) and grouped by string equality — so on a part whose runs mostly differ, every string was built only to be found unequal. The rule is now a pairwise predicate comparing run properties as **trees**. Three of the simplifier's five passes are additionally skipped when a read-only scan shows nothing in the part can trigger them: those three are pure structural rebuilds outside the elements they target, so their absence proves the pass is an identity. The other two keep running unconditionally, on purpose — coalescing renormalizes a run's `w:t` and separating run children drops run attributes, so they are **not** identities even when they merge and split nothing, and a guard there would change output rather than only cost. Measured through the editor's own incremental block render: `refresh()` **88 ms → 73 ms** on a 101-run paragraph. Every DOCX→HTML conversion gets it, including the per-keystroke block re-render.

**How the grouping change is guarded.** A grouping rule that changes quietly merges a little more or a little less and still emits well-formed markup, so the rule it replaced is kept as an executable oracle. `CoalesceGroupingTests` requires the two to agree on every adjacent pair in every run container of the committed corpus — paragraphs, hyperlinks, content-control bodies, and the tracked insertions and deletions where the two could most plausibly diverge, since those keyed on author, date and id as well as properties — with floors so a shrunken corpus cannot make the assertion vacuous.

**One disclosure about how that reached the branch.** Those two library files were edited in the working tree and swept into an earlier commit by a `git add -A` whose message described only a Playwright spec fix. They were not reviewed or tested at that point. The follow-up commit is the correction: read, tested against the full .NET suite, documented, measured, and with the dead public API it had left behind (`HasCoalescableRuns`, unreachable from any caller) removed and `CanCoalesceAdjacent` made internal.

**The converter no longer emits a redundant direction marker.** Splitting exact-line-height content around a break used to append a zero-metric LTR marker span after every break, even though the paragraph's declared LTR direction already supplies that information — one extra DOM span per line. It is now omitted only in that provably redundant compact LTR case; bidi exact-line paragraphs keep their marker, with a focused regression on both branches.

## The controls are ordinary document text

The keyboard map used to be a side panel drawn on the framebuffer grid, which rendered it about seven pixels high: the right words at an unusable size. It is now **four short bold 18pt document paragraphs** above the screen — 24px in the fitted desktop page, still at least 14px at the tested 60% embed scale — kept out of every frame conversion by the context fence. Four fixed lines rather than one, because a single oversized OOXML run can clip.

## What this cartridge can no longer do — and why the raycaster stays

The text cartridges re-read their world from the paragraph on every resume: type a wall into the map, resume, and walk into it. Doom cannot. Its level is BSP geometry in a WebAssembly heap, and nothing painted into an image means anything to it. doomgeneric's host interface (DrawFrame, GetKey, SetTitle, sound) never lets level state cross the boundary in either direction, and Doom's precomputed BSP could not accept a typed wall even if it did.

That round trip is **dropped for this cartridge rather than faked**. Pause, edit, undo, copy, paste and save all still work, and the next frame replaces the image. The trick itself is not lost — it is what The Docx Dungeon plays, and what the restored **Freedoom E1M1 cartridge** plays on a real Doom-format level: its 1,175 rasterized linedefs are readable, typable document text, which is precisely the demo the real engine cannot be. `?cart=e1m1` selects it; `?cart=doom` — the default on both host pages — selects the engine.

## Nothing binary lands in the repository

The engine (3 MB) and the IWAD (10 MB gzipped) are **not** in this repo. They would show up in every clone forever and never diff usefully, so both are pinned jsDelivr URLs; `docs/demo/vendor/` carries only license text and a provenance record, 48 KB in total.

| | Pin | Verified |
|---|---|---|
| Engine | `gh/grubbyplaya/doomgenericjs@99d7a55…` | 200, 3,097,628 B, `application/javascript`, CORS `*`, SHA-256 `efd7c347…` |
| IWAD | `gh/JSv4/freedoom-iwad@70ee6ec9…` | 200, 10,278,537 B, CORS `*`, inflates to SHA-256 `7323bcc1…` |

Both are pinned by 40-hex **commit**, which is stronger provenance than a tag or a vendored copy — jsDelivr cannot serve different bytes for a given SHA, and unlike a tag it cannot be moved under a version already published. The engine's digest is identical to the file that was briefly vendored here; the IWAD's is identical to Freedoom's own release asset.

**Why the IWAD needs a repo of its own:** GitHub serves release assets *without* CORS, so a browser cannot fetch Freedoom's release directly, and jsDelivr cannot bridge it either because Freedoom's WADs are build artifacts and are not in its tree. Hence [`JSv4/freedoom-iwad`](https://github.com/JSv4/freedoom-iwad), which exists only to be CDN-addressable and carries `COPYING.txt` verbatim beside the file — BSD-3-Clause requires the notice travel with a redistribution, and that repo is one.

**The tests deliberately do not use the CDN.** A browser suite depending on a CDN being up *and* on a sibling repo being reachable would fail for reasons unrelated to the change under test. `npm/scripts/fetch-doom-iwad.mjs` pulls the release asset server-side (no CORS server-side), verifies the recorded SHA-256, and gzips it into the Playwright webroot; the specs load it same-origin through the cartridge's own URL gate, exercising the same `DecompressionStream` path a visitor takes.

## Licensing — the one non-MIT file in the repository

`docs/demo/doom-cart.js` is combined at runtime with a GPL engine, so **that single file is offered under GPL-2.0-or-later** and says so in its header. Everything else stays MIT, including `ascii-arcade.js`, which imports it: MIT is GPL-compatible, so each file remains separately available under its own terms, and only the combination a browser assembles when the Doom cartridge is selected is GPL-2.0. The engine sits behind a dynamic `import()`, so a visitor who plays the other cartridges fetches neither artifact.

**On the IWAD.** doomgenericjs carries id's shareware `doom1.wad` in its tree, so it is already CDN-addressable and a third the size — this PR does not use it. id's shareware terms cover redistributing the complete, unmodified shareware *package*; building a product's default experience on an IWAD lifted out of it is not that, and it is copyrighted commercial game content either way. Freedoom exists precisely so a project can ship a playable Doom without that question. doomgeneric's own IWAD table recognises `freedoom1.wad` and treats it as a retail Doom, so the filename is the entire configuration.

## Security

`?wad=` points the cartridge at an IWAD you host and are licensed to play, and must be **same-origin**. There is deliberately **no engine override**: `import()` executes whatever it fetches, on this page's origin and with its privileges, so a URL parameter naming the module would be remote code execution from a crafted link. The URL gate is an allowlist of `{our own pin} ∪ {same-origin}` — a pin is trusted because it is written in the source, not because of where it points.

An earlier revision of this branch did expose such a parameter (`?doomEngine=`); it is gone, and a regression test asserts a cross-origin `?wad=` lands in the cartridge's error state with no request reaching the foreign host.

Sound effects are wired through WebAudio from Doom's own DMX lumps, every entry point guarded so a refused or suspended AudioContext costs the player sound and never the game. Music is declined at the driver, which `i_jsmusic.c` supports directly.

## Validation

`npm/tests/demo-arcade-doom.spec.ts` proves, for the engine cartridge:

- **It is really Doom.** The engine reads the IWAD's identity out of the file and reports `Freedoom: Phase 1` through `DG_SetWindowTitle`; nothing but the engine produces that string. Doom's own frame counter is then required to keep advancing.
- **The frame is in the DOCUMENT, losslessly.** The canvas block holds exactly one native image at its exact 320 × 200 natural size, on the editor surface — not a canvas mounted beside the document.
- **It sustains the design point.** The browser regression measures completed image mutations *and* decoded, animation-frame-presented sources across five seconds; both must stay at or above ten FPS. Measuring only mutations would pass a game that never reached the screen.
- **The controls stay readable.** All four control lines present and non-overflowing, 24px in source and at least 14px scaled, outside the framebuffer block — so the readability fix cannot silently regress at either size.
- **The keyboard reaches Doom.** Driven through Doom's own menu into a level, the 3-D view is near-static with no key held and swings when a turn key is, while the status bar stays put. The static control is the load-bearing half: a recorded attract demo drives the player and would never sit that still.
- **It is still a document.** Pause, save, reopen: a real ZIP, and the reopened block carries the same image. Copy and paste duplicates it as a second native image; Undo and Redo scrub frames.
- **The conversion speedup changes no output.** The full .NET suite passes and the warning baselines are unmoved — library 132, tests 775, both measured with `--no-incremental`. `CoalesceGroupingTests` pins the grouping equivalence against the rule it replaced.

The restored `npm/tests/demo-arcade-freedoom.spec.ts` proves the E1M1 cartridge with its BFS autopilot (navigate the real geometry to a pickup, save the frame it was taken on), and the headless clear-the-level autopilot now runs **both** raycaster packs.

The branch was brought level with `main` during review, with the full .NET suite run on each merged tree before pushing, because this branch changes run coalescing and `main`'s concurrent work touched revision markup: no shared files, but they meet at runtime.

Four things CI caught on this branch, all fixed: a mobile controls test that still read raycaster-only `player.dx` after being repointed at Doom; a liveness assertion that demanded 60 frames inside 60 s, which measured throughput while claiming to measure liveness (cut to 10, which a frozen engine still cannot produce); a hardcoded 26 rendered line boxes in the mobile grid check, which read a denser grid as a line fold and now measures each cartridge against its own row count; and a landing-page guard that asserted the canvas *text* contains a bezel, which the native-image screen empties by design.

## Follow-up merged into this branch (#639)

A line-by-line review of this PR was merged back in as #639, in two commits:

**Review gaps closed.** The library changes, the oracle test and the Playwright specs held up under adversarial reading; what needed fixing was documentation and dead code. `CHANGELOG.md` had been left describing the branch's superseded character-grid implementations (including a `P` projection toggle and `?projection=` parameter that no longer exist) — its `[Unreleased]` Doom narrative now describes the shipped native-image design, and the load-bearing incremental-image library fix gained the `### Fixed` entry it was missing. Dead code in `doom-cart.js` (`mergeBezelsIntoPicture`, an unused `frameHash` helper whose comment claimed specs use it, a write-only `edited` flag) is gone, along with a stale "vendored beside this page" comment in `arcade.html`.

**E1M1 restored as cartridge 3.** `docs/demo/freedoom-e1m1.js`, `tools/wad2cart.mjs` + its test, and `demo-arcade-freedoom.spec.ts` return byte-identical from `main` (wad2cart's test re-proves the checked-in level is the converter's canonical output); `FREEDOOM_PACK` and `freedoomCart()` return to `ascii-arcade.js` with the four control lines the new controls paragraphs expect; every demo suite covers all four cartridges again. This PR's earlier **Removed** section no longer applies: nothing is removed except the intermediate character projections of the engine's framebuffer, whose budgeted painters survive as exported helpers for the headless renderer tests.

The arcade opens on DOOM by default on both host pages; `?cart=quest|dungeon|e1m1` pick the other three cartridges explicitly.


JSv4:
## CodeQL is red and I can't finish it from here — I need one thing from someone with security-tab access

**Failing check:** `CodeQL` — "New alerts in code changed by this pull request: 1 critical", on both `ab064fd` and `e041a23`.

I can read the check's *summary* but not the alert. This environment has no `gh` CLI and no code-scanning MCP tool, and the alerts API needs write-scope auth, so I can see that one critical alert exists but not its rule ID, file, or line. Everything below is what I could establish without it.

### What I already fixed (a real bug, but apparently not this alert)

`arcade.html` and `index.html` were reading `?doomEngine=` from the query string and passing it into a dynamic `import()`. `import()` executes what it fetches, on this origin, so a crafted link to the Pages site was arbitrary script execution there. Fixed in `e041a23`: the engine override is gone from both pages, and `loadEngine()` now refuses any cross-origin engine or IWAD URL, with a regression test that a cross-origin `?wad=` errors without a request leaving the page.

That was worth fixing on its own merits, but the alert count stayed at exactly **1 critical** across both heads — so it looks like this was never the alert.

### The two candidates I can see

**1. The vendored engine (most likely — it's the only *new* critical-severity sink in the diff).**

`docs/demo/vendor/doomgeneric/doomgeneric_module.js:4643`, in Emscripten's Node-environment branch:

```js
var cp = require('child_process');
var ret = cp.spawnSync(cmdstr, [], {shell:true, stdio:'inherit'});
```

That's a textbook `js/command-line-injection` (security-severity 9.8). It is dead code in a browser — Emscripten's `system()` shim for a Node host that this page never takes — and it's 3 MB of upstream build output vendored byte-for-byte, so it isn't patchable here without diverging from the hash recorded in `docs/demo/vendor/NOTICE.md`.

**2. A pre-existing alert re-attributed to this PR.**

`arcade.html:107` → `:144` does `params.get("engine")` → `import(engineUrl.href)` — the same code-injection shape, and genuinely exploitable. But it is **not mine**: it's on `main` today, in all six demo pages (`app`, `arcade`, `golf`, `index`, `observatory`, `player`), and `git diff origin/main..HEAD -- docs/demo/arcade.html` touches none of those lines. CodeQL's own summary warns for exactly this: *"Alerts not introduced by this pull request might have been detected because the code changes were too large."*

### Why I stopped instead of pushing a third guess

The right fix for (1) is a `paths-ignore` on `docs/demo/vendor/**` — vendored third-party build output shouldn't be scanned, which is also why it's already marked `linguist-vendored` in `docs/demo/vendor/.gitattributes`. **But this repo has no CodeQL workflow file** — analysis runs through *default setup*, where path filters are a repository setting, not something I can commit. Converting to an advanced `codeql.yml` would work, but it disables default setup and I'd have to re-create the whole csharp/python/actions/javascript matrix by hand; getting that subtly wrong reduces scanning coverage repo-wide, off a Doom PR. That's not a call I should make unilaterally.

The right fix for (2) is to stop letting a URL parameter choose a module across all six pages — the same gate I just added to the cartridge. That's a deliberate, documented convention for pinning the library version, and reworking it belongs in its own PR, not this one.

### What would unblock me

Either of these and I'll push the fix straight away:

- **The alert's rule ID and location** (from `/security/code-scanning?query=pr%3A620`) — paste it here and I'll act on it.
- **Or a decision on the two proposals**: add `docs/demo/vendor/**` to the default-setup path filters (Settings → Code security → CodeQL → Edit configuration), and/or tell me to open a separate PR gating `?engine=`/`?games=`/`?dock=` to same-origin across the demo pages.

I haven't re-run the check: this is a deterministic analysis result, not a flake, so a re-run would return the same thing.

The rest of CI on `e041a23` is unaffected by this, and I'm still watching the PR — I'll pick it back up the moment there's an alert to act on.

---

JSv4:
## The 13 MB is out of the repo — and that may settle the CodeQL question by itself

`dfe8c6f` removes both vendored binaries. **92,175 lines and ~13 MB out of the tree**; `docs/demo/vendor/` is now 48 KB of license text and provenance.

**Engine** → pinned to `grubbyplaya/doomgenericjs@99d7a55` via jsDelivr's `gh` endpoint. Verified directly: HTTP 200, 3,097,628 bytes, `content-type: application/javascript`, `access-control-allow-origin: *`, immutable cache, and a SHA-256 **identical** to the file that was vendored. Pinning by 40-hex commit is stronger provenance than a vendored copy, not weaker — it's upstream's own published build served from upstream's own tree.

**IWAD** → pinned by tag to a sibling repo. It needs one because GitHub serves release assets *without* CORS, so a browser can't fetch Freedoom's release directly, and jsDelivr can't bridge it either since Freedoom's WADs are build artifacts and aren't in its tree.

### On the CodeQL alert

This very likely resolves it without anyone having to dig out the alert. The prime suspect I named above was `doomgeneric_module.js:4643` — Emscripten's Node-only `cp.spawnSync(cmdstr, [], {shell:true})`, a textbook `js/command-line-injection` at 9.8. **That file is no longer in the tree**, so if it was the alert, it's gone.

If CodeQL is still red on `dfe8c6f`, then it was candidate 2 — the pre-existing `params.get("engine")` → `import()` in the demo pages, which is on `main` in all six of them and which this PR doesn't touch. That would be worth its own PR, and I'd be happy to open it: the fix is the same allowlist gate I put on the cartridge, applied to `?engine=`/`?games=`/`?dock=` across the pages.

Either way the ambiguity resolves on the next run, so my earlier ask is probably moot — please ignore it unless CodeQL stays red.

### ⚠️ Merge blocker

**`JSv4/freedoom-iwad@v0.13.0` does not exist yet.** Until it's created and tagged, the shipped pages will show the cartridge's own "could not start" screen. `docs/demo/vendor/NOTICE.md` has the exact recipe; it's about four commands:

```bash
curl -LO https://github.com/freedoom/freedoom/releases/download/v0.13.0/freedoom-0.13.0.zip
unzip -j freedoom-0.13.0.zip freedoom-0.13.0/freedoom1.wad freedoom-0.13.0/COPYING.txt
sha256sum freedoom1.wad   # must be 7323bcc1…
gzip -9 freedoom1.wad
# commit freedoom1.wad.gz + COPYING.txt, then: git tag v0.13.0 && git push origin v0.13.0
```

The `COPYING.txt` matters — BSD-3-Clause requires the notice travel with a redistribution, and that repo *is* one. Say the word and I'll create and populate it.

### The tests deliberately don't use the CDN

A browser suite that depended on a CDN being up *and* on a sibling repo having been tagged would fail for reasons unrelated to the change under test. `npm/scripts/fetch-doom-iwad.mjs` pulls the release asset server-side (no CORS server-side), verifies the recorded SHA-256, and gzips it into the Playwright webroot; the specs load it same-origin through the cartridge's own URL gate, exercising the same `DecompressionStream` path a visitor takes. Dependency-free zip reader, cached by existence, so a local `npm test` pays one download once.

So **the Doom specs pass or fail on their own merits right now**, regardless of the blocker above.

### Security note

The URL gate is now an allowlist of `{our own pin} ∪ {same-origin}` rather than same-origin alone — a pin is trusted because it's written in the source, not because of where it points. `?wad=` still must be same-origin; there's still no engine override at all.

Re-verified in Chromium through the guard with absolute URLs: engine imported, IWAD fetched and inflated, Doom playing at 22.7 turning against 0.38 standing, status bar 0. The literal cross-origin hop couldn't be exercised in my container — its egress proxy resets the browser's connections — so jsDelivr's headers, content-type and digest were measured directly instead.

---

JSv4:
⚠️ **The last two heads on this branch have not been checked by CI.** The PR shows green, but on 5 checks instead of the usual 14.

**What is missing.** `ci.yml` (`build-and-test`, `build-matrix`, `python-tests`, `build-npm`), the Playwright `test` job, and `claude-review` were not created for `6c1157c` (14:16Z) or `b92eaa5` (14:28Z). The last `ci.yml` run on this branch is `8038463` at 12:00Z. CodeQL and the four `Analyze` jobs still run, which is why the checks list looks healthy.

**It is not a repo-wide outage and not a config change.** `ci.yml` ran normally on other branches at 13:15Z, 13:27Z, 13:28Z and 14:07Z — after this branch stopped triggering it. Nothing recent touched `.github/workflows/`, the trigger is still `pull_request: branches: [main]`, and the PR is open, not draft, based on `main`. I could not determine the cause, and I have no way to fix it from here: there is no queued or failed run to re-run, and `ci.yml` has no `workflow_dispatch` trigger, so there is nothing to dispatch either.

**What I ran locally instead**, on `6c1157c` (which is `b92eaa5` plus a changelog-only commit):

| check | result |
|---|---|
| Full .NET suite | 4,223 passed, 0 failed, 3 skipped |
| `demo-arcade-doom.spec.ts` (chromium) | 8 passed |
| `demo-arcade-mobile.spec.ts` (chromium-pixel5) | 7 passed |
| arcade / observatory / ascii-animation / social | 22 passed |
| `pretest`: `tsc --noEmit`, demo-logic suites, package boundary | clean |

The browser specs ran against a freshly built wasm bundle, so they cover the `WmlToHtmlConverter` direction-marker change in `6c1157c` as it actually reaches the browser. `demo-arcade-mobile.spec.ts` was run under `chromium-pixel5` explicitly — it is the only project that spec runs in, and the default `chromium` project silently skips it.

That is not a substitute for CI: it is one machine, one OS, and it does not cover the `build-matrix` targets (macOS, Windows), the Python tests, or `build-npm`. **Please re-run CI before merging** rather than reading the green tick as coverage.

---

JSv4:
✅ **Resolved — CI is running again and all 14 checks pass on `c70c00c`.** My earlier warning about missing checks no longer applies; please disregard it.

The gap was transient, not a broken trigger. Pushing the merge of `main` at 15:53Z produced a full run: `build-and-test`, `test` (the 20-minute Playwright job), all three `build-matrix` targets, both `python-tests`, `build-npm`, `claude-review`, CodeQL and the four `Analyze` jobs — every one green. The two heads that were skipped (`6c1157c`, `b92eaa5`) were never checked by CI, but everything in them is contained in this head, which has been.

The merge was worth doing on its own terms: `main` had landed four commits, two of them in `RevisionProcessor.cs` and `NoteReferenceOps.cs` (a recorded note's definition carrying insertion markup, and accepting a redline taking a wholly-deleted note's definition with it). This branch rewrites the run-coalescing rule in `MarkupSimplifier`, so revision markup is exactly where the two could interact — no shared files, but they meet at runtime. Full .NET suite on the merged tree before pushing: **4,230 passed, 0 failed, 3 skipped**. `CHANGELOG.md` conflicted, both sides appending under `[Unreleased] → Changed`; every entry was kept.

The branch is level with `main` and has no open review threads.

---

Issue #563, opened 2026-08-26
Inline and margin comment modes emit marker links to targets those modes never render


JSv4:
## Gap

`ProcessCommentReference` always emits the comment marker as a link:

```html
<a href="#comment-1" id="comment-ref-1" class="comment-marker">[1]</a>
```

The `#comment-N` target only exists in **EndnoteStyle** mode, where `RenderCommentsSection` renders `li id="comment-N"`. In **Inline** mode no comments section exists at all, and in **Margin** mode the pagination engine strips ids from the cloned margin notes (`makeClonedMarginCommentInert`), so in both modes every comment marker is a link to nowhere.

## How it surfaces

The standalone paginated export validates fragment targets in `normalizeFragmentTargets` (`npm/src/export-browser.ts`). A dangling marker becomes:

- under `unsupportedContent: "warn"` (default): an inert, `aria-disabled` label plus a `fragment_target_unavailable` warning — the current, quietly-degraded behavior;
- under `unsupportedContent: "strict"`: a closed `resource_policy_failure` in `running_story_placement` — **any commented document fails a strict inline/margin export**.

Until #540 this was partly masked: a document with comment replies or resolved comments failed strict earlier, in `package_preflight`, on the (now retired) comment-topology warnings. With those families drawn and the warnings gone, the dangling marker is the first thing strict trips over. The `draws comment topology rather than warning about it` spec deliberately asserts strict under `endnotes` for exactly this reason.

## Where a fix would go

`Docxodus/WmlToHtmlConverter.cs`, `ProcessCommentReference`: only emit `href="#comment-{id}"` when the render mode actually produces that target (EndnoteStyle). For Inline/Margin the marker should be an anchor-free `<span>`/`<a>` carrying the same id, classes and metadata (`data-comment-id`, tooltip), so nothing downstream has to repair a link the converter knew was dangling. The export-side inert-link downgrade stays as defense in depth for genuinely broken documents.

Found while closing #540.

PR #678, opened 2026-09-02
Emit a note backlink only when its target is in the output

### LINKED ISSUE #667: Converter emits dangling backlinks for uncited endnote definitions
## Summary

`WmlToHtmlConverter` emits a clickable backlink for an endnote definition even when the document contains no matching `w:endnoteReference`. The generated `href` therefore targets an id that cannot exist, and standalone export reports a `fragment_target_unavailable` warning for markup the converter authored itself.

## Reproduction

The in-repo fixture `TestFiles/NVCA-Model-COI.docx` contains an uncited, blank `w:endnote` with `w:id="1"` in `word/endnotes.xml` and no corresponding endnote reference in a document story.

Render it with endnotes enabled:

```csharp
var html = WmlToHtmlConverter.ConvertToHtml(
    new WmlDocument("NVCA-Model-COI.docx"),
    new WmlToHtmlConverterSettings { RenderFootnotesAndEndnotes = true });
```

The endnotes section contains:

```html
<li id="en-1" value="1">
  <p> <a href="#en-ref-1" class="en-backref">↩</a></p>
</li>
```

No element with `id="en-ref-1"` exists because no `w:endnoteReference` exists. The Node export path consequently records:

```text
fragment_target_unavailable: Fragment target #en-ref-1 is unavailable in the final page tree.
```

Under strict unsupported-content policy, a converter-authored dangling fragment can make export fail rather than merely warn.

## Expected behavior

An uncited note definition may be retained and rendered for package fidelity, but it must not receive a hyperlink to a nonexistent reference. The converter should either:

- omit the backlink for definitions with no matching reference; or
- render an inert backlink-shaped label without `href`.

This is analogous to the target-aware marker behavior fixed for comment modes in #563.

## Acceptance criteria

- A note backlink is emitted only when the output contains its target id.
- A fixture with an uncited footnote and an uncited endnote produces no dangling fragment links.
- Cited notes retain working forward and backward navigation.
- Standalone export of `TestFiles/NVCA-Model-COI.docx` no longer reports `#en-ref-1` as unavailable.
- Warn and strict export-policy tests cover this case.

JSv4:
## The bug

`WmlToHtmlConverter` gave every footnote and endnote definition it rendered a return-arrow hyperlink back to the reference that cites it — including definitions that nothing cites.

`TestFiles/NVCA-Model-COI.docx` is the in-repo case: `word/endnotes.xml` carries a blank `w:endnote` with `w:id="1"`, and no `w:endnoteReference` for it exists in any story. Word keeps orphans like this and the converter renders them, which is the right call for package fidelity. But the arrow it rendered alongside them pointed at `#en-ref-1`, an id the conversion never emitted. Standalone export then reported

```
fragment_target_unavailable: Fragment target #en-ref-1 is unavailable in the final page tree.
```

for markup the converter had authored itself. Under a strict unsupported-content policy that is not a warning, it is a failed export.

## Why it happened

The two halves of a note's navigation come from different code paths reading different sources. The marker (`#fn-N`) is emitted from a `w:footnoteReference` in a rendered story; the return arrow (`#fn-ref-N`) is emitted from a `w:footnote` definition in `footnotes.xml`. Neither half checks that the other exists, and the relationship fails in both directions — an orphaned definition has no reference, and a reference can outlive its definition.

## The fix

One post-pass, `RemoveUnresolvableNoteLinks`, alongside the two whole-tree passes the converter already runs after the transform.

**It resolves against the finished tree, not against the numbering tracker.** That matters. `FootnoteNumberingTracker` only scans the main document part, so a reference living in a header or footer never reaches it — a tracker-based check would have stripped a backlink whose target genuinely does exist. Asking "is this id in the output?" is the question the acceptance criterion actually states, and it is story-agnostic by construction.

**The two directions are repaired differently**, because they are not the same kind of thing:

- A **return arrow** is navigation chrome. An unresolvable one is removed outright, along with the separator space rendered with it, so an orphaned note keeps no trailing gap.
- A **marker** carries the note's visible number, which is document content. An unresolvable one keeps its text and loses only its dead `href`, becoming the inert label the issue's second option describes.

Only note links are touched (`#fn-…`, `#en-…`). Dangling cross-reference or bookmark links are a different question with a different right answer, and comment backlinks were already made target-aware in #563.

## Validation

- On the NVCA fixture, the conversion goes from one dangling fragment target (`en-ref-1`) to none, while all 94 cited footnotes keep working navigation in both directions.
- Six new tests in `WmlToHtmlConverterNoteLinkTests` cover: the orphaned endnote still renders but without an arrow; the whole NVCA conversion has zero dangling fragment targets; cited footnotes keep both directions; a fixture carrying *both* an uncited footnote and an uncited endnote produces no dangling links; and the mirror case — a cited note whose definition was deleted — keeps its marker text but loses its href.
- **Non-vacuous:** with the post-pass disabled, five of the six fail. The sixth asserts the fixture really does carry an uncited endnote, so the others cannot pass by testing nothing.
- Full .NET suite: 3931 passed, 0 failed, 3 skipped.

## Scope note

The issue's last two acceptance bullets concern `@docxodus/export`. This change removes the *cause* — the converter no longer authors a link to a target it did not emit — so the export warning cannot be raised for this document. The export path's own handling of a genuinely dangling link (strip the href, mark it `aria-disabled`, warn or fail per policy) is unchanged and still correct for links it did not author.

Closes #667

https://claude.ai/code/session_01VSnwkK1Nx6zZb2RnnoxKdx

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,20 @@ jobs:
           dotnet build tools/docx2html/docx2html.csproj -c Release --no-restore
           dotnet build tools/docx2oc/docx2oc.csproj -c Release --no-restore
 
+      # These projects are deliberately outside Docxodus.sln, so nothing else compiles
+      # them. Without this step a breaking change to the library rots them silently
+      # (issue #662: the benchmark harness sat un-buildable on main for a week).
+      - name: Build out-of-solution tools and benchmarks
+        run: |
+          dotnet build benchmarks/complex-form-doc/ComplexFormBenchmark.csproj -c Release
+          dotnet build benchmarks/docxdiff-stress/DocxDiffStress.csproj -c Release
+          dotnet build tools/delivery/delivery.csproj -c Release
+          dotnet build tools/diffharness/diffharness.csproj -c Release
+          dotnet build tools/manifest-fuzz/afl/aflharness.csproj -c Release
+          dotnet build tools/manifest-fuzz/fuzzer/fuzzer.csproj -c Release
+          dotnet build tools/manifest-fuzz/replay/replay.csproj -c Release
+          dotnet build tools/screenshots/redline/redline-screenshot.csproj -c Release
+
       - name: Run tests
         env:
           DOCXODUS_EVAL_ARTIFACTS: ${{ github.workspace }}/eval-artifacts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- `WmlToHtmlConverter` no longer emits a return-arrow hyperlink for a footnote or endnote
+  definition that no story cites. Word keeps such orphaned definitions and the converter still
+  renders them for package fidelity, but the arrow used to target an id nothing had emitted, so
+  standalone export reported `fragment_target_unavailable` for markup the converter authored
+  itself — enough to fail an export under a strict unsupported-content policy. Note navigation is
+  now resolved against the finished output: an unresolvable return arrow is dropped, and the
+  mirror case — a cited note whose definition is missing from the part — keeps its visible marker
+  but loses its dead `href` rather than becoming another dangling link.
+
 - `OpenXmlRegex.Match` no longer throws a `NullReferenceException` when called without a callback
   on PowerPoint/DrawingML content (the `a:p`/`p:p` namespaced path) — it now returns the match
   count like the Wordprocessing path already did.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@ All notable changes to this project will be documented in this file.
   now resolved against the finished output: an unresolvable return arrow is dropped, and the
   mirror case — a cited note whose definition is missing from the part — keeps its visible marker
   but loses its dead `href` rather than becoming another dangling link.
+- The complex-form benchmark harness (`benchmarks/complex-form-doc`) compiles again. Removing the
+  legacy comparison engine deleted the harness's terminal summary along with its legacy stage,
+  leaving the `int`-returning entry point with a path that never returned (`CS0161`). The summary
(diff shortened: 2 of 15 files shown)


JSv4:
### Follow-up: the first version of this fix hung the editor

CI caught it, and it was a real defect in my change, not a flake: `npm/tests/editor-reconcile.spec.ts` — *reconciled edits survive save/reopen losslessly* — timed out at `insertFootnote`, on all three retries, and reproduced locally every time. `main` runs the same spec in 15s.

**The mistake.** The sweep asks *"is this note link's target id in the output?"* That is a fair question of a complete document and the wrong question of a fragment. The editor's incremental renderer builds a shell whose body holds only the blocks it asked for, so a citing paragraph rendered on its own has no notes section to point at — and the sweep stripped the `href` off its citation marker on every re-render.

**How it was localized.** Calling `RenderBlocksHtmlForReview` straight through the bridge with the same anchors returned in 6–33 ms and handed back the marker as `<a id="fn-ref-1" class="footnote-ref"><sup>1</sup></a>` — no `href`. So the render was never slow; the missing `href` was what sent the reconciler into a spin that blocked the browser's main thread outright (a `waitForFunction` for a plain variable could not run either). The equivalent .NET call had always returned in ~0.6 s, which is what made this look WASM-specific at first.

**The fix.** A new internal `WmlToHtmlConverterSettings.RendersDocumentFragment`, set by `BuildBlockConverterSettings` — the single owner of the block-render profile, and already home to `SkipFormattingPartsSimplification`, the existing "this is the incremental path, not the full one" knob. Completeness checks over the finished tree are skipped when it is set. Nothing else changes, and the full-document render — the one the reported export bug lives on — is unaffected.

**Verification**
- `editor-reconcile.spec.ts`: 7 passed in 15.0s, matching `main`'s timing.
- Full .NET suite: 3932 passed, 0 failed, 3 skipped.
- New regression test renders a citing block through the incremental path and asserts the marker keeps both its class and its `href`. Flipping `RendersDocumentFragment` back to `false` fails it, so it pins this bug specifically rather than the general sweep.

PR #729, opened 2026-09-06
feat(history): validate a portable document history graph

JSv4:
## Why

A portable history file must contain one document's complete referenced history, not a dump of a shared store. Its reader must reject broken history before import can publish anything.

## Change

- Add an internal, bounded graph reader rooted at one captured head. Preserve exact snapshot/record identities, retry receipts, effects, restore branches, and operation proposals; never enumerate storage.
- Validate ancestry, receipt retention, both radix branches, actual endpoint deltas, and original/mapped text intent. Keep large payloads out of the metadata cache.
- Exercise real NVCA and legal-contract documents, isolated reopening/replay, missing/corrupt data, forged edges, resource limits, and shared-graph scaling.

This is the first foundation PR. ZIP files, safe writable import, document-scoped APIs, client bindings, concise frontend docs, and committed archive artifacts follow in the stack. No GUI or transport.

Validation establishes referential and recorded-content integrity, not authentication or an independent audit of the backend's acceptance/conflict policy. Recorded outcomes are preserved, not re-adjudicated.

## Validation

Focused graph and affected snapshot/codec/journal/update regressions: 50 tests passed. Adversarial 5.6-sol review drove revisions for proposal integrity, retained-branch compatibility, receipt loss, forked commit tips, duplicate resolutions, and peak-memory bounds.

PR #730, opened 2026-09-06
Portable history files: export and read-only open

JSv4:
## Why

Let users carry a document's version history in one `.docxhistory` file, without exposing the application's storage directory.

## What

- Stream one captured head and its reachable blobs into a versioned ZIP; preserve snapshot bytes and IDs.
- Open independently through a read-only, document-scoped API with strict inventory, integrity and resource checks.
- Include real agreement/charter archives, exact DOCX versions, a conflict proposal and comparison files as compatibility fixtures.

No transport or writable import in this slice. Stacked on #729; import and client controls follow.

## Validation

22 archive/file-artifact tests, including malformed input, limits, ZIP64, concurrent reads and concurrent publication. Adversarial GPT-5.6-sol review; revisions recorded in the review comment.


JSv4:
Adversarial GPT-5.6-sol review completed at `6b90888`: no P1/P2 blockers remain.

Revisions: clarified the independent graph-read and ZIP-metadata budgets; added counted-input checks proving byte/count rejection before directory traversal. Preserved separate compressed-snapshot and raw-payload limits.

Reviewed pinned export, ZIP/ZIP64 preflight, exact inventory, stream ownership/concurrency, readonly capabilities, hostile input and real compatibility artifacts. Final regression run: 22/22 passing. Raw outer ZIP serialization is not an identity guarantee; immutable blob and snapshot bytes are.

PR #731, opened 2026-09-06
Portable history files: exact import and document API

JSv4:
## Why

Resume editing a portable history without losing its original versions or retry receipts, or overwriting a local document's history.

## What

- Validate in isolation, copy immutable blobs, then atomically initialize the exact captured head only if absent.
- Add optional initialization to memory/filesystem head stores; ordinary CAS still increments by one.
- Bind native reads/checkpoint/restore to one document ID, with a short usage guide.

Stacked on #730. No transport, GUI or implicit saves.

## Validation

34 import/head/process tests pass: real agreement and charter files; exhaustive before/after blob-copy faults; races; missing/corrupt storage; lost acknowledgements; actual process kills before/after publication and fresh-process recovery. Clean build: zero errors. Adversarial GPT-5.6-sol review recorded separately.


JSv4:
Adversarial GPT-5.6-sol review: no P1/P2 blockers remain at `2a435dd`.

Reviewed isolated validation, optional capability, atomic absent-only initialization with ordinary CAS, exact retry/repair/corruption, collisions/races, commit-point cancellation and stream ownership. Clarified that caller-owned input stays open while the archive reader and entry leases close before publication.

Final gate: 34/34 native import/head/process tests pass, including exhaustive real-file copy faults and actual process kills before/after publication. Clean build: zero errors.

PR #732, opened 2026-09-07
feat(history): portable-file controls for npm, Python, and MCP

JSv4:
## Summary
- Add scoped history controls and readonly `.docxhistory` opening in npm/Python, plus MCP file export/import. Exact-identity import uses optional atomic head initialization.
- Expose retained decisions/proposals and arbitrary-pair redlines, preserving the existing raw native comparison API.
- Add a [concise frontend guide](https://github.com/JSv4/Docxodus/blob/feat/history-file-clients/docs/history-controls.md), conservative byte-client limits, and real DOCX/archive integration tests. No transport, GUI, or autosave.
- **Raises the WASM wire-size guardrail from 5 MB to 5.25 MB** — see below.

Portable-history group 4; #729–#731 are merged. Rebased onto current main (`2a1e6f9`, which now carries #721 and #728).

## The size guardrail moves, and why

Reaching the `.docxhistory` reader/writer from the browser costs **46 KB** of brotli wire
payload. Before this change the WASM history surface was
`read`/`updates`/`create`/`list`/`get`/`export`/`materialize`/`replay`/`resolveTime`/`restore`
only; `exportArchive`, `importArchive`, `exportDocx`, `compare`, `operations`, `getOperation`
and `exportOperationProposal` pull ZIP entry walking, manifest and graph validation, and the
change-set codec into the bundle for the first time.

The margin for that was already gone before this branch. Measured, clean builds:

| Tree | Brotli wire size |
|---|---|
| main before the dense-text AOT profile coverage | 4981 KB |
| main after it | 5105 KB (15 KB under the old 5120 KB gate) |
| main + this branch | 5151 KB — **31 KB over** |

Re-recording the AOT profile without the dense-text workload was measured as the alternative:
it recovers 28 KB (5123 KB), which is *still over the old gate*, and it costs the
formatting-template and batched-identity coverage that workload exists to provide —
`wasm-packaging.md` documents that coverage as deliberate and explicitly not game-specific.
Paying 31 KB of wire for a feature that already ships in every other transport was the better
trade, so `scripts/build-wasm.sh` moves to 5.25 MB, leaving ~225 KB of headroom.
`docs/architecture/wasm-packaging.md` records the measurements, both options, and the reasoning.

## Verification
- Full CI green on `8ef6177`: **14/14 checks**, including `build-npm`, the browser suite, both
  Python versions, and all three `build-matrix` platforms.
- WASM budget verified with a **clean** AOT rebuild (`bin`/`obj` removed so the recorded profile
  is not served from cache; 40 assemblies recompiled): **5154 KB against the 5376 KB gate**.
  An incremental rebuild after a profile change silently reuses cached AOT output and reports a
  stale number — worth knowing if you re-measure this.
- Full `npm run build` and `npx tsc --noEmit` both complete.
- Two assertions in the new coverage were tightened after review: both `compare` tests asserted
  only that the returned bytes parsed into a `DocxSession` (a constructor cannot return null), so
  a comparison that silently returned an unchanged package would have passed. They now require
  `ListRevisions()` to be non-empty. The readonly binding's `exportArchive` result was checked
  only for non-null bytes; it now reopens those bytes through `OpenArchiveAsync` and requires the
  same `DocxHistoryArchiveInfo` back.
- Adversarial GPT-5.6-sol review/revision: no remaining P1/P2 findings. Follow-up #733 verifies
  cross-runtime artifact handoff and CI retention.

Note: the warning-baseline figure quoted in the original description (187/804) predates #728;
main is now at 204/829, and #734 takes it to 175/671. This PR does not change it.

---

https://claude.ai/code/session_017H3myrxq9faz9LoKvGcgwm


JSv4:
Adversarial **GPT-5.6-sol** review completed against current main; final verdict: **no remaining P1/P2 findings**.

Revisions made: expose operation decisions/proposals and comparisons across clients; bound expanded archive work; reject oversized/unknown MCP import metadata; preserve the existing raw native comparison API while adding one product-policy owner; make scoped document IDs readonly; align concise API examples.

Evidence: native 50/50, Chromium 12/12, Python 8/8; full npm/WASM build and size budget pass; non-incremental native build 0 errors with unchanged warning baseline. Reviewer performed independent read-only code/design review; execution results came from the implementation session.

PR #738, opened 2026-09-07
feat(history): durable checkpoint commands and browser storage

JSv4:
A lost save acknowledgement can leave an app unsure whether a checkpoint was published. Add explicit checkpoint commands that persist complete requests before publication and recover the original result across reloads. The optional IndexedDB adapter serializes head initialization/publication and keeps one durable pending request per document.

Stack 1/3: #738 → #739 → #740 (416 added lines here). The following PRs add the panel and runnable editor example.

Validation: TypeScript build and source/test typecheck; three Chromium scenarios cover a real agreement recovered after reload and a later save, stale drafts and append-only restore, and independent database connections competing for heads and requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Durable browser checkpoint commands: `HistoryCheckpoints` persists a complete save/restore
+  request through a host-supplied `HistoryCheckpointJournal` *before* publishing it, so a lost
+  acknowledgement no longer leaves an app unsure whether a checkpoint exists. `retry()` replays
+  the stored request and returns the original result — the exact view the first attempt would
+  have produced — even when another editor has since published a newer version. The journal is
+  insert-only per document: a second tab's `put` returns the request already there rather than
+  replacing it, which is what makes a competing draft fail loudly instead of overwriting a
+  publication whose outcome is still unknown. Only a genuine `StaleHead` (the compare-and-swap
+  provably did not apply) clears a pending request.
+- `openIndexedDbHistoryStore(name)`: optional browser-local `HistoryStorage` plus a per-document
+  `HistoryCheckpointJournal`, backed by IndexedDB. Head initialization and compare-and-swap share
+  one read/write transaction on the same object store, so independent connections — separate tabs
+  included — serialize against each other; stored bytes are SHA-256 verified against their
+  reference before they are written. Nothing installs it: no storage, autosave or retention policy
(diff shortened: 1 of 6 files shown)


JSv4:
## Adversarial review

I went through the checkpoint state machine and the IndexedDB adapter line by line, and traced the failure modes each is meant to survive. **No correctness defects found.** The parts I specifically tried to break and could not:

- **Defensive copying is real, not decorative.** `save()` reaches `structuredClone` synchronously — an async function body runs to its first `await`, and neither `save`, `start` nor `exclusive` awaits before the clone — so the existing reload test mutating `bytes`/`metadata` immediately after the call genuinely proves the captured request is immune. Same for `publish()` re-cloning before `journal.put`.
- **Discarding a pending request on `StaleHead` is safe.** It looked like a hole ("never clear an uncertain publication"), but a lost acknowledgement replays through the core's request journal and returns the original view rather than a CAS failure — which the reload test proves by publishing a competing version first. So `StaleHead` really does mean the publication did not happen.
- **`advanceHead` computes the next revision from `expected` rather than `current`.** Different from the in-memory reference, but equivalent: the CAS gate only proceeds when the two are equal, and both fall back to `'0'` when absent.
- **The IndexedDB transaction helper propagates failures.** An exception thrown from a request callback aborts the transaction, so `onabort` rejects rather than the promise resolving `undefined`.

### What I changed

Two gaps, both about what *isn't* exercised rather than what's wrong:

1. **The guards protecting an uncertain publication had no coverage.** All three scenarios drive happy paths plus one lost acknowledgement. Added tests for opening over a pending request belonging to a different document, retrying with nothing pending, and issuing a second command while one is in flight — the last also asserting the refused command neither published a second version nor disturbed the one running.

2. **The cross-tab race the design exists for was untested end to end.** Scenario 3 tests the journal's insert-only `put` directly, but nothing exercised `HistoryCheckpoints` reacting to it. Added a test where a competing tab records its request *after* this tab opened its controls: the local save is refused, `Retry` publishes the competing tab's original request, and the assertions prove the local draft never reached storage — one version exists, carrying the other tab's label and exactly its bytes.

Both use a small in-memory journal so they exercise the protocol rather than IndexedDB, which the storage-level scenarios already cover.

3. **No `CHANGELOG.md` entry**, which the repo's own feature workflow requires and this adds substantial public API. Added one under `[Unreleased]` for the checkpoint commands and the optional IndexedDB store.

### Not changed, worth knowing

`crypto.randomUUID` and `crypto.subtle` restrict this to secure contexts, so a host serving over plain HTTP on a LAN gets a runtime failure rather than a graceful one. That is consistent with the rest of the history stack and #739 documents the requirement in `npm/README.md`, so I left it.

---

PR #739, opened 2026-09-07
feat(history): accessible version and collaboration controls

JSv4:
Apps can mount a document-history panel to browse saved versions, preview or download a selected DOCX, compare arbitrary versions, save checkpoints, restore, and explicitly share a portable history file. Read-only archive readers use the same panel. The editor remains host-owned, and conflicting controls are disabled during calls.

Stack 2/3: #738 → #739 → #740, based on #738 (452 added lines here). Includes time lookup, collaboration conflict status, retry feedback, and usage documentation.

Validation: TypeScript build/typecheck and three Chromium panel scenarios using the real agreement and charter archives. Tests verify paginated browsing, exact DOCX/proposal downloads, tracked-change comparison, standalone archive reopening, lost restore acknowledgement recovery, and phone-width layout. The checkpoint tests also pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@ All notable changes to this project will be documented in this file.
   replacing it, which is what makes a competing draft fail loudly instead of overwriting a
   publication whose outcome is still unknown. Only a genuine `StaleHead` (the compare-and-swap
   provably did not apply) clears a pending request.
+- `mountHistoryControls(container, options)`: an accessible, responsive document-history panel an
+  app mounts beside its own editor. It pages through version metadata, previews or downloads a
+  selected DOCX, compares any two versions as tracked changes, finds a version by time, lists
+  recorded collaboration with its resolved conflicts, and — when given `checkpoints` and a
+  `capture` callback — saves, restores and retries checkpoints. The panel never owns, replaces or
+  closes the host's editor or reader: previews go to a host callback so browsing cannot discard a
+  draft, `destroy()` awaits the command still in flight, and controls that would conflict with a
+  running call are disabled while it runs. Read-only `.docxhistory` archives mount the same panel
+  by omitting `checkpoints`. `historyControlError(error, pending)` is exported separately so a host
+  can render the same plain-language explanations outside the panel.
 - `openIndexedDbHistoryStore(name)`: optional browser-local `HistoryStorage` plus a per-document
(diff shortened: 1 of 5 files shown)


JSv4:
## Adversarial review

Went through the panel's lifecycle, its disabled-state logic and every DOM path that touches version metadata. **No correctness defects found.** Specifically checked and cleared:

- **Version metadata is never interpolated as markup.** Labels, authors and messages all go through `textContent`, including the `<option>` text and its `title`. The new keyboard/time test in #740 pins this with a ``'&lt;img src=x onerror=...&gt;'`` label.
- **`destroy()` genuinely drains.** It flags first, aborts the listener signal, removes the element, then awaits the in-flight command; `run()`'s `finally` and `preview()`/`download()` all re-check `destroyed`, so a slow export that lands after teardown cannot repaint into a detached panel or hand bytes to a host that has torn down its viewer.
- **The `Busy` guard has no window.** `run()` reads and writes `this.active` synchronously around `Promise.resolve().then(action)`, so a second click cannot slip past it.
- **`compare` cannot read `undefined`.** The only way `before.value`/`versions.value` are empty is an empty record set, and `hasVersion` already disables the button in that case.
- **A pinned older view is intentional, not a paging bug.** Listing from `checkpoints.view.version.id` after an idempotent retry hides newer versions, and the status text after a retry says exactly that ("Refresh history to check for newer versions").

### What I changed

Two things the scenarios could not reach, both about coverage rather than behaviour:

1. **The two guards that run before a panel exists were untested.** Added a test for a `pageSize` outside 1–100 (including a fractional one) and for a `checkpoints` instance bound to a different document than the `reader`. It also asserts a refused mount leaves the host's container untouched — the property that lets a host retry without cleaning up first.

2. **`historyControlError` had no coverage at all**, despite being exported for hosts to use outside the panel. Added a test pinning each recognised code to its own explanation, an unrecognised code still surfacing the underlying message, a pending checkpoint taking precedence over an unrecognised code, and a thrown non-`Error` value degrading to a usable sentence instead of `undefined`. The last assertion pins that the explanations stay distinct, so a future code cannot silently collapse into an existing one.

3. **No `CHANGELOG.md` entry.** Added one under `[Unreleased]` for the panel and the exported error mapper.

Also merged the #738 branch forward so this branch carries its changelog and tests.

---

PR #740, opened 2026-09-07
feat(history): safe portable-history editor example

### LINKED ISSUE #672: Add durable document version history and non-destructive restore
## Objective

Add durable, host-controlled document version history and restore semantics to the browser/headless surfaces.

This should build on immutable DOCX snapshots and existing semantic diff/redline capabilities rather than treating editor undo state as durable history.

## Proposed capability

- Create a named/versioned snapshot of the current canonical DOCX bytes.
- List snapshot metadata: id, parent id, document digest, author, timestamp, label/message, and optional application metadata.
- Retrieve or export an exact snapshot.
- Produce a semantic change summary or native redline between any two snapshots.
- Restore an earlier snapshot by creating a new head version; never silently erase the intervening history.
- Let the host own storage through an adapter rather than embedding a cloud dependency.
- Expose the same primitives to the browser editor, npm headless API, and agent-facing integrations.

## Design questions

1. Whether operation journals are retained in addition to canonical snapshot bytes.
2. Snapshot cadence and compaction policy for large documents.
3. Binding snapshots to renderer/PageMap/delivery artifacts without conflating those derived outputs with source identity.
4. How revision authors and application actors map into version metadata.
5. How restore interacts with open sessions, optimistic revisions, and collaborative clients.

## Acceptance criteria

- A host can plug in memory, filesystem, object-store, or database persistence through one documented interface.
- Snapshot identity is content-bound and detects corrupted or mismatched bytes.
- Restore is non-destructive and creates an auditable new version.
- Diff/redline between versions uses the existing `DocxDiff` owner and retains accept/reject invariants.
- Version metadata round-trips without modifying the DOCX unless the host explicitly requests embedded provenance.
- Browser UI provides history browsing, change inspection, and restore confirmation as an opt-in surface.
- Headless consumers can use the complete capability without mounting the editor.

JSv4:
Add a runnable TypeScript editor example with opt-in IndexedDB checkpoints, read-only portable history, import/resume with embedded document identity, and version previews separate from the draft. Conflicting imports, stale saves, and invalid uploads preserve the open editor. Durable retries retain the original captured inputs and receipt.

Restore now asks for confirmation naming the selected version and explaining that the draft and later versions are kept. Cancel leaves the published head, versions, and durable retry journal unchanged. Confirmed restores append an auditable checkpoint. Acknowledged checkpoint callbacks clear discard warnings only for the confirmed capture; newer edits stay protected.

Also preserve captured heads when mounting or resuming imports, add paginated collaboration decisions/proposal downloads, and fix ribbon opening when the host hides file actions.

Stack 3/3: #738 → #739 → #740, based on #739. This PR has 572 additions and 31 deletions (603 changed lines). Merge the stack in order; this final PR closes the issue when merged into `main`.

Validation:
- TypeScript build/typecheck, including the example; example bundle and package boundary audit.
- 29 Chromium history/editor/ribbon tests, including keyboard restore cancellation, exact retry recovery, preserved later versions, stale drafts, uploads, and portable-file reopening.
- 38 native history/storage/client/MCP tests. Real agreement and charter redlines now explicitly verify that reject reproduces the earlier version and accept reproduces the later version.
- 8 Python headless history tests, including process restart and exact exports.

Closes #672

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,12 +25,34 @@ All notable changes to this project will be documented in this file.
   running call are disabled while it runs. Read-only `.docxhistory` archives mount the same panel
   by omitting `checkpoints`. `historyControlError(error, pending)` is exported separately so a host
   can render the same plain-language explanations outside the panel.
+- `mountHistoryControls` gained `onCheckpoint(view, action)`, which reports an acknowledged save,
+  restore or retry before the version list reloads. A host uses it to decide whether its open draft
+  is now saved: because a retry can acknowledge a *request captured earlier* — possibly by another
+  tab — only a fresh save whose capture is still the current draft may clear an unsaved-work
+  warning. Recorded collaboration also pages instead of rendering every operation at once, and each
+  entry can expand into its decision detail. `HistoryCheckpoints.open` takes an optional third
+  argument, an already captured view, so an editor opened at an exact version keeps that head
+  paired with the bytes on screen instead of silently adopting the latest.
+- A runnable TypeScript editor example (`npm/examples/history.ts`, built with
+  `npm run build:history-example`) pairing the ribbon editor, IndexedDB checkpoints and the history
+  panel: portable `.docxhistory` files open read-only, importing their embedded identity resumes
+  editing, previews render separately from the draft, and a pending save reopens its captured draft
+  after a reload.
 - `openIndexedDbHistoryStore(name)`: optional browser-local `HistoryStorage` plus a per-document
   `HistoryCheckpointJournal`, backed by IndexedDB. Head initialization and compare-and-swap share
   one read/write transaction on the same object store, so independent connections — separate tabs
   included — serialize against each other; stored bytes are SHA-256 verified against their
   reference before they are written. Nothing installs it: no storage, autosave or retention policy
   is created by using the history API, and clearing browser site data removes the store.
+
+### Fixed
+
+- `mountRibbon({ fileActions: false })` threw `Docxodus ribbon: template is missing "save"` on the
+  first `open()`. That option removes the quick group holding New/Open/Save, but opening a document
+  re-enabled the Save button unconditionally, so every host that hides the file actions — precisely
+  the hosts that supply their own persistence — could mount the ribbon and never open anything in
+  it. The rest of the surface already looked its controls up optionally; this was the last
+  unconditional lookup.
 
 ## [12.2.0] - 2026-09-07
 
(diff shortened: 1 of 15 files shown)


JSv4:
## Adversarial review

Reviewed the example's draft-safety logic, the paged activity list, the `onCheckpoint` contract and the ribbon fix. **No correctness defects found.** The things I tried hardest to break:

- **The `dirty` bookkeeping is right for the case it exists for.** `capture()` stamps `capturedGeneration = editGeneration`, and `onCheckpoint` only clears the discard warning when the two still match — so an edit that lands *while the save is in flight* keeps the warning. Unconditionally resetting `capturedGeneration = -1` afterwards is also correct: a later `retry` can only ever acknowledge a request captured earlier, possibly by another tab, so it must not clear anything.
- **Size is checked before bytes are read.** `file.size > MAX_HISTORY_ARCHIVE_BYTES` precedes `arrayBuffer()`, and the test proves it by making `arrayBuffer` throw.
- **Every failing path mounts its replacement before tearing down the incumbent.** `installDraft`, `openFile` and `back` all build the candidate ribbon/panel, `await controls.ready`, and only then destroy the current one — so a bad file leaves a working editor, which is what the malformed-upload test observes.
- **The activity pager's arithmetic is sound at both ends.** `slice(max(0, length - shown - pageSize), length - shown).reverse()` is exact with an empty operation list (`hidden` becomes true immediately) and on the final partial page.
- **The changed order in `historyControlError` is defensible.** `pending` now outranks `ResourceLimit`/`UnsupportedVersion`/`InvalidManifest`. It only affects panel-originated failures, and there an unconfirmed checkpoint is the more urgent thing to say; `StaleHead` and `ImportConflict` still keep their own instruction because they already tell the user what to do next. I've pinned that precedence in a test rather than changing it.

### On the ribbon fix

`mountRibbon({ fileActions: false })` removing `[data-dxr-files]` while `open()` did `require("save")` was a real crash for **every** host that hides the file actions, not just this example — those are exactly the hosts supplying their own persistence. I confirmed `save` is the only removable control still looked up unconditionally (`new` and `file` already used `control()`), so the fix is complete.

### What I changed

1. **Added a direct regression test for the ribbon fix.** It was only caught end to end, through the example. `ribbon-surface.spec.ts` now mounts with the file group hidden, opens a document, and asserts it renders editable blocks in its own WASM session while Save is genuinely absent and the rest of the ribbon still wires up. That fails if the unconditional lookup ever returns, without depending on the example's plumbing.

2. **Extended the error-explanation test** to `UnsupportedVersion` and `InvalidManifest` and to the new precedence described above.

3. **No `CHANGELOG.md` entry.** Added entries for `onCheckpoint`, the paged activity list, the optional captured-view argument to `HistoryCheckpoints.open`, the example itself — and the ribbon crash under `### Fixed`, since it is a user-facing bug fix in shipped API rather than an example detail.

Also merged #739 forward, plus your `fix(history): confirm restores` commit.

### Not changed, worth knowing

The restore confirmation uses `window.confirm`, which a sandboxed iframe without `allow-modals` silently resolves to `false`. The panel degrades gracefully there (status reads "Restore canceled"), and a host in that situation can call `checkpoints.restore()` directly as the docs say — so this is a documented limit rather than a defect, but it is worth a line in `docs/history-controls.md` if embedding is a target.

---

